PR #33, opened 2021-05-28
fix encoder base64url to base64 to avoid bouncycastle DecoderException

jorgesanchezperez:
- Bouncycastle expect base64 encoded instead of base64Url.
- bouncycastle.util.encoders raise a DecoderException when `x5c` and `key` are encoded using base64Url ('-' and '_' not valid in base64)
     - `org.bouncycastle.util.encoders.DecoderException: unable to decode base64 string: invalid characters encountered in base64 data`


jorgesanchezperez:
Closing the PR to be consistent with encoding of JWK set.
https://www.rfc-editor.org/rfc/rfc7638

PR #37, opened 2021-06-10
67: Use consent prefixes according to type

christian-brindley:
Resolves: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/67

### DIFF
--- a/config/7.0/obdemo-bank/ig/scripts/groovy/RCSGetConsent.groovy
+++ b/config/7.0/obdemo-bank/ig/scripts/groovy/RCSGetConsent.groovy
@@ -1,110 +1,99 @@
 import groovy.json.JsonOutput
 
-/*
- * getAccountList
- *
- * Stub for account list from resource server
- * Return dummy data
- */
-
 def getAccountList(username) {
 
     return [
-      [
-        "currency" : "GBP",
-        "account" : [ [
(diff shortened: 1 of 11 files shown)


config/7.0/obdemo-bank/ig/scripts/groovy/RCSGetConsent.groovy
```diff
@@ -1,110 +1,99 @@
...
+
+    def accountList = []
+
+    accountData.accountDatas.forEach( data -> {
+            accountList.push (      [
+                    "currency" : data.account.Currency,
+                    "account" : [ [
+                                          "currency" : null,
+                                          "account" : null,
+                                          "identification" : data.account.Account[0].Identification,
```
matt-wills, line 71: Will `data.account` or `data.account.Account` ever be null?

matt-wills (approved):

config/7.0/obdemo-bank/ig/scripts/groovy/RCSGetConsent.groovy
christian-brindley: They shouldn't be - but in any case, this is disposable code, which will be replaced in its entirety by the actual RCS service.

PR #45, opened 2021-07-16
Change the field pemCert to x5c

jorgesanchezperez:
- Changed the field `pemCert` to `x5c` in the response json to align with the standard
- Script: `JwkmIssueCert.groovy`
Issue: https://github.com/SecureBankingAccessToolkit/securebanking-openbanking-uk-functional-tests/issues/2

### DIFF
--- a/config/7.0/obdemo-bank/ig/scripts/groovy/JwkmsIssueCert.groovy
+++ b/config/7.0/obdemo-bank/ig/scripts/groovy/JwkmsIssueCert.groovy
@@ -175,8 +175,8 @@ if(publicKey instanceof RSAPublicKey){
 
 Response response = new Response(Status.OK)
 response.getHeaders().add("Content-Type","application/json");
-//response.setEntity(pemCert + pemKey);
-logger.debug("Final JSON " + JWKJson)
-response.setEntity(JWKJson)
+
+logger.debug("Final JSON " + JWKJsonResponse)
+response.setEntity(JWKJsonResponse)
 
 return response
--- a/config/7.0/obdemo-bank/ig/scripts/groovy/JwkmsIssueCert.groovy
+++ b/config/7.0/obdemo-bank/ig/scripts/groovy/JwkmsIssueCert.groovy
@@ -154,25 +154,13 @@ X509Certificate issuedCert  = new JcaX509CertificateConverter().setProvider(BC_P
 issuedCert.verify(caCertificate.getPublicKey(), BC_PROVIDER);
 
 // Build response with cert and private key
-
-/*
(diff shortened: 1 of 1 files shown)


christian-brindley:
As we're doing this, I guess it would be best to stick to https://datatracker.ietf.org/doc/html/rfc7517#page-9 - i.e.

  The "x5c" (X.509 certificate chain) parameter contains a chain of one
   or more PKIX certificates [RFC5280].  The certificate chain is
   represented as a JSON array of certificate value strings.  Each
   string in the array is a base64-encoded (Section 4 of [RFC4648] --
   not base64url-encoded) DER [ITU.X690.1994] PKIX certificate value.

So we should actually be using base64 encoding rather than base64url encoding.

jorgesanchezperez:
> As we're doing this, I guess it would be best to stick to https://datatracker.ietf.org/doc/html/rfc7517#page-9 - i.e.
> 
> The "x5c" (X.509 certificate chain) parameter contains a chain of one
> or more PKIX certificates [RFC5280]. The certificate chain is
> represented as a JSON array of certificate value strings. Each
> string in the array is a base64-encoded (Section 4 of [RFC4648] --
> not base64url-encoded) DER [ITU.X690.1994] PKIX certificate value.
> 
> So we should actually be using base64 encoding rather than base64url encoding.

DONE!

config/7.0/obdemo-bank/ig/scripts/groovy/JwkmsIssueCert.groovy
```diff
@@ -150,37 +155,28 @@ X509CertificateHolder issuedCertHolder = issuedCertBuilder.build(csrContentSigne
...
+}
 
 Response response = new Response(Status.OK)
 response.getHeaders().add("Content-Type","application/json");
-//response.setEntity(pemCert + pemKey);
-logger.debug("Final JSON " + responseJson)
-response.setEntity(responseJson)
+
+logger.debug("Final JSON " + JWKJsonResponse)
+response.setEntity(JWKJsonResponse)
```
jorgesanchezperez, line 180: The response must be similar to:
```json
{
  "kty": "RSA",
  "x5t#S256": "gEingApbsMdYq5pRMoaAcQ5Mq55TflsYFpasId8omVg",
  "e": "AQAB",
  "kid": "1477206499341455272",
  "x5c": [
    "MIIKYTCCCEmgAwIBAgII......vQl"
}
```

jorgesanchezperez: It will be possible compute the private key with fields `e`, `d` and `n`

christian-brindley (approved):

PR #82, opened 2022-01-25
200: Create temporary IG route to return RS discovery JSON document with already implemented endpoints

andra-racovita:
Issue:
https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/200

* remove '$' from the condition of the route 08-ob-account-access.json


andra-racovita:
Invalid pull request

PR #83, opened 2022-01-26
Temporary RS discovery route

andra-racovita:
For deployment on andra-racovita env


BohoCode:
You can turn this into a single commit using 

`git rebase -i HEAD~10`

You can then squash all those commits into a single commit and the push it to this PR using `git push -f`. It will be neater to have a single commit for your change.

BohoCode (approved):
Looks good to me!

PR #146, opened 2022-07-15
376: international payment funds confirmation route

jorgesanchezperez:
- Updated route names for domestic payment
- Added route for international payment funds confirmation
Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/376


jorgesanchezperez:
Invalid PR, We need to design a solution for exchange rate information and aspsp checks if necessary.

PR #157, opened 2022-08-17
484: Calculate elements mandatory header

jorgesanchezperez:
- CalculateResponseElementsInRS script creates and set the 'x-fapi-financial-id' header  when necessary
Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/484


jorgesanchezperez:
Invalid PR, confusion of terms `interaction-id` and `financial-id`, the tpp needs to send that header.

PR #158, opened 2022-08-18
443: Replace nimbus.jose load method with an asynchronous http call for loading jwk_keys

frlaura-jianu:
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/443

Updated the filter processing to use only async methods. The keys are retrieved using an async http call to the open banking directory jwks_uri, and once the response is received the correct key is identified. A promise with the key is created and when it's completed, the RSAKey is avaialble.


jorgesanchezperez (approved):

jorgesanchezperez:
Good Job @frlaura-jianu and @mariantiris

PR #159, opened 2022-08-18
Adding necessary objects on the request for funds confirmation

mariantiris:
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/412


BohoCode (changes requested):
How do you intend to publish document the format of the string representing the intent that the back office API must recieve? 

Customers implementing production (rather than sandbox) OB systems will need to implement these APIs and passing the JSON string around will make it very hard for the to implement. 

In my opinion we should be passing the OB generated consent object in this API rather than a string called IntentJsonObject?

dbadham-fr:
Closing this, work implemented in another PR.

PR #161, opened 2022-08-26
497: (First approach) create extension gateway modules to convert intents to OB data model objects

jorgesanchezperez:
- Created business logic gateway conversion module
- Created filter conversion module with openig dependencies
- Created parent pom for modules
- Implementation of conversion logic
- Implementation of conversion filter
- Created unit tests coverage for all modules
- Added assembly maven plugin to build filter library with dependencies and copy it into config/7.0/obdemo-bank/ig/lib to be deployed in IG when is build

    Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/498


securebanking-openbanking-uk-gateway-conversions/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/converters/GenericIntentConverter.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * Generic converter to provide a common way conversion between json string and OB object types
+ * @param <T> represents OB object
+ * @param <S> represents a json string
+ */
+public class GenericIntentConverter<T, S> {
```
dbadham-fr, line 34: Will S ever be anything other than a java.lang.String?

If not then we can remove the S param
Otherwise if we keep S then I think it is more readable to make the sig GenericIntentConvert<S, T> as we are mapping S -> T.

jorgesanchezperez: Yes, I will delete the para S because always will be a java.lang.String, it was a inheritance of first approach we not needed anymore.

dbadham-fr:
For my understanding, how does the IDM representation of an intent differ from the OB json representation (generally speaking)?

In the Account Access example, it seems that the conversion is to use jackson to decode the string, I assume this is just dropping any additional fields in the IDM intent object which are not found in the OB object? Are there any other common conversion patterns which we need to do?

securebanking-openbanking-uk-gateway-conversions/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/converters/AccountAccessIntentConverter.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class AccountAccessIntentConverter extends GenericIntentConverter<OBReadConsentResponse1, String> {
+
+    private static final Logger logger = LoggerFactory.getLogger(AccountAccessIntentConverter.class);
+
+    public AccountAccessIntentConverter() {
+        super(AccountAccessIntentConverter::convertToOBObject);
+    }
+
+    private static OBReadConsentResponse1 convertToOBObject(String jsonString) {
```
dbadham-fr, line 31: I'd make this an instance method and use the objectMapper field in the base class. Otherwise this is creating a new mapper obj each time it is invoked.

Alternatively, instead of passing a conversion Function to the base class, it could have an abstract convertToObObject method which subclasses (AccountAccessIntentConverter etc) would implement.

dbadham-fr: Actually, probably just need to review the class hierarchy a little. The base class has a mapper field but it never uses it itself.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,110 @@
...
+ *  }
+ *  }
+ * </pre>
+ */
+public class IntentConverterFilter implements Filter {
+
+    private static final Logger logger = LoggerFactory.getLogger(IntentConverterFilter.class);
+
+    private final IntentType intentType;
+    private final String intentContent;
```
dbadham-fr, line 65: What is the use case for passing the intentContent string as config? Won't we always want to take the intent from the http request/response body?

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,110 @@
...
+                entity = request.getEntity().getString();
+            }
+            Object converted = ConverterFactory.getConverter(intentType).convertFromJsonString(entity);
+            request.setEntity(converted);
+            // add the entity to the response
+            return next.handle(context, request)
+                    .thenOnResult(response -> {
+                        response.setEntity(converted);
+                        response.setStatus(Status.OK);
+                    });
```
dbadham-fr, line 88: I think the response conversion needs some thought, the current impl will send the converted intent attached to the request (or from the config) in the response. However I would assume that the response may be a different object from the request? 

We may only want to support converting the request or response payload (but not both), as the response may already be a valid OB object. For instance, when calling into the back office API to calculate charges for a payment, the request object will start off as an IDM json and then be converted to OB obj to send to the back office, I assume the back office then replies with an OB object (so no conversion on the response path is required).

Please correct me if I have misunderstood the requirements / how the impl works...

securebanking-openbanking-uk-gateway-conversions/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/converters/AccountAccessIntentConverter.java
jorgesanchezperez: Deleted `mapper` field from generic class, to be keep a static method to get the generic converter mapper to be used statically in tests and in the static methods of class hierarchy.
A hierarchy class can get the mapper using the static method `generiConverterMapper()` and add new configurations if necessary before call the `readValue(...)` mapper method to map the content.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: No always we want to get the intent from response body, could exist a case that we want to convert only part of the IDM intent content, as example: `request.getEntity().getJson().<key>.<key>.toString()` to convert only part of the intent entity or the case that we have the intent content in a variable or property that we can inject in the Filter using the configuration.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Cases covered in this solution:
- A case when the TPP request the consent pushed previously (to check the status for example, means GET the Consent), where will not necessary call the RS backend to build the response, the TPP will retrieve a consistent OB data model object from the response.
- A case where we need to build the OB data model object to call RS, in that case we can define in the API request body and in the `ResponseEntity` as an OB data model object instead of a String as is defined currently:
```java
// currently
ResponseEntity calculateElements(
            @ApiParam(value = "Create an Account Access Consents", required = true)
            @Valid
            @RequestBody String body,
.....
// Future (calculate elements and update the ob object with calculated elements to return it to IG and update the intent on IDM)
ResponseEntity<OBReadConsentResponse1> calculateElements(
            @ApiParam(value = "Create an Account Access Consents", required = true)
            @Valid
            @RequestBody OBReadConsentResponse1 body,
```
this will generate a more consistent swagger documentation for the customer if they want to implement his own RS.
It means also that we need to modify the RS backend properly.

securebanking-openbanking-uk-gateway-conversions/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/converters/AccountAccessIntentConverter.java
jorgesanchezperez: Deleted mapper field from generic Converter class and moved to a Generic mapper singleton class.

jorgesanchezperez:
Dependencies used on gateway conversion business logic module have been compiled with `jdk 14` (byte code 58), IG only supports byte code classes 55 (`jdk 11`).
- If the uk gateway modules are compiled with **jdk14**: 
- ```java.lang.UnsupportedClassVersionError: com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter has been compiled by a more recent version of the Java Runtime (class file version 58.0), this version of the Java Runtime only recognizes class file versions up to 55.0```
- If the uk gateway modules are compiled with **jdk11** (changing jdk and language level): `cannot access com.xxx, bad class file ... class file has wrong version 58.0, should be 55.0`

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
dbadham-fr: The use cases make sense, I am just concerned with the logic. 

I am not familiar with IG filters so perhaps I am misunderstanding how this new filter will be called, when this filter is added to an existing chain then I should be able to understand the flow better.

My concern is with lines 86/87 in the snippet, when the filter is called as part of a chain, how do we know that we always want to set the response entity and status to the converted request entity and 200 response? Can't filters further into the chain potentially have mutated (or produced an entirely different response object) or have raised an error?

For example, given a filter chain:
- request -> FilterA -> IntentConverterFilter -> FilterB -> SomeHandler
- SomeHandler -> FilterB -> IntentConverterFilter -> FilterA-> response

Then if FilterB or SomeHandler produce/edit the response entity (and/or status code), then the IntentConverterFilter will clobber these values.

Therefore we may want to have some configuration around when the filter is applied (either through different classes or some flag), so that we can control if the converter logic is applied to request only, responses only or both.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Deleted that logic, now the configuration will have a field to indicate where the filter needs to get the incoming payload.
```json
{
    "name": "IntentConverterFilter-accessAccountConsent"
     "type": "IntentConverterFilter",
    "config": {
        "intentType": "ACCOUNT_ACCESS_CONSENT",
        "payloadFrom":"REQUEST",
        "resultTo": ["RESPONSE"]
    }
}
```

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: New field `resultTo` introduced to indicate where will be set the conversion result object (request, response or both)
```json
{
    "name": "IntentConverterFilter-accessAccountConsent"
     "type": "IntentConverterFilter",
    "config": {
        "intentType": "ACCOUNT_ACCESS_CONSENT",
        "payloadFrom":"REQUEST",
        "resultTo": ["RESPONSE"]
    }
}
```

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,211 @@
...
+                    .as(enumConstant(MessageType.class));
+            logger.trace("payload From {}", payloadFrom);
+            List<MessageType> payloadTo = config.get(CONFIG_FIELD_RESULT_TO)
+                    .defaultTo(List.of(MessageType.RESPONSE.toString()))
+                    .as(messageTypeList());
+            return new IntentConverterFilter(intentType, payloadFrom, payloadTo);
+        }
+    }
+
+    private static Function<JsonValue, List<MessageType>, JsonValueException> messageTypeList() {
```
dbadham-fr, line 150: I think this is only called by the Heaplet, if so then move it inside the Heaplet class.

As the Heaplet logic is getting larger, I think it would make the IntentConverterFilter more readable if the Heaplet is converted from an inner class to a top level class in its own file

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,211 @@
...
+                resultList.add(MessageType.valueOf(value));
+            }
+            return resultList;
+        };
+    }
+
+    private String getEntity(Context context, Request request, Handler next) throws IOException, InterruptedException {
+        logger.trace("Payload from {}", this.payloadFrom);
+        if (payloadFrom.equals(MessageType.RESPONSE)) {
+            return getEntity(next.handle(context, request).getOrThrow());
```
dbadham-fr, line 170: Promise.getOrThrow is a blocking call, which is not allowed unfortunately, see: https://backstage.forgerock.com/docs/ig/7.2/gateway-guide/extending.html#about-custom-extensions

dbadham-fr: I don't think this method is needed, you can create a new method which has the logic to go from a Message obj to the converted json byte[].

This new method can then be plugged into the request flow and the response flow (within the promise callback within method processResult)

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Configuration changed:
```json
{
  "name": "IntentConverterFilter-AccountAccessConsent",
  "type": "IntentConverterFilter",
  "config": {
    "messageType": "request",
    "entity": "#{request.entity.string}"
    "resultTo": [
      "response"
    ]
  }
}
```
## Configuration

| field         | description                                                                                   | required | default         |
|---------------|-----------------------------------------------------------------------------------------------|----------|-----------------|
| `messageType` | The type of message for which to convert the entity                                           | Yes      | N/A             |
| `entity`      | Expression<String> to be replaced for a jsonPayload content to be converted in string format. | No       | [not replaced.] |
| `resultTo`    | Indicates where will set the conversion result                                                | No       | `REQUEST`       |

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Refactor done.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,287 @@
...
+            entity.evalAsync(bindings)
+                    .then(result -> {
+                        if (result == null) {
+                            throw new Exception("Unexpected null result from this expression " + entity);
+                        }
+                        return result;
+                    }, neverThrown()).thenOnResult(result -> {
+                        content.set(result);
+                    });
+            return content.get();
```
dbadham-fr, line 225: this will only work if the evalAsync determines that it can evaluate the expression without an async call i.e. behaves the same as calling the eval synchronous function.

if evalAsync has to do an async call then content.get() will mostly likely return null (as the async callback thenOnResult won't have executed to call content.set yet).

So you need to determine if the eval ever needs to do an async call, if it does not then you can swap evalAsync for eval which returns a value. Otherwise you need to restructure the code to do the relevant processing in the promise callbacks / chain more callbacks.

The eval method logs an error if it thinks you should be using the async version, so this can help with investigation:
```
 public T eval(final Bindings bindings) {
        if (valueExpression.isDeferred()) {
            logger.error("Trying to invoke synchronously an expression ('{}') that requires async", original);
        }
        return evaluateNow(bindings);
    }
```

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @VisibleForTesting
+    IntentConverterFilter(MessageType messageType, final Expression<String> entity) {
+        this(messageType, entity, List.of(MessageType.REQUEST));
+    }
+
+    @VisibleForTesting
+    IntentConverterFilter(MessageType messageType, final List<MessageType> resultTo) {
+        this(messageType, null, resultTo);
+    }
```
dbadham-fr, line 127: these constructors don't add any functionality which can't be achieved through the public constructor, therefore I think it would be cleaner to create factory methods under src/test for these rather than telescoping the constructor solely for testing purposes.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Yes, we no need the async method. change done.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Constructors for test deleted, created a factory test filter to get the instance of filter.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,261 @@
...
+
+    @Override
+    public Promise<Response, NeverThrowsException> filter(Context context, Request request, Handler next) {
+        if (logger.isInfoEnabled()) {
+            printFilterInfo();
+        }
+
+        if (messageType.equals(MessageType.REQUEST)) {
+            return next.handle(context, request).thenAsync(response -> processRequest(request, response, bindings(context, request)));
+        }
```
dbadham-fr, line 121: I think the request needs converting before it is handed on to the rest of the chain, doing work in the thenAsync is on the response chain which means the request will have been passed on to the other filters (and to the RS) unconverted.

Something along the lines of:
```
if (messageType.equals(MessageType.REQUEST)) {
            processRequest(request, null, bindings(context, request));
            return next.handle(context, request);
        }
```

I think the processRequest method sig needs reviewing, the request should not be concerned with the response, the filter should have the following 3 modes:

-  Convert requests only, do the conversion on the request entity and hand the updated entity on to the next handler in the chain
- Converts response only, do the conversion on the response entity and hand the updated entity on to the next handler
- Convert Request entity and Response entity (use both pieces of logic from above)

jorgesanchezperez:
Closed, to be refactor in a new PR

PR #169, opened 2022-09-07
Adding cnfKey verification for all required routes.

mariantiris:
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/238
Description: Implementing the certificate thumbrint validation for all the routes where it was missing. This is required for certificate bound access tokens.

Filter documentation:
https://backstage.forgerock.com/docs/ig/7.1/reference/CertificateThumbprintFilter.html

### DIFF
--- a/config/7.0/obdemo-bank/ig/routes/routes-service/11-ob-account-access.json
+++ b/config/7.0/obdemo-bank/ig/routes/routes-service/11-ob-account-access.json
@@ -81,6 +81,21 @@
           }
         },
         {
+          "comment": "Extract client certificate thumbprint for cert bound access tokens",
+          "name": "CertificateThumbprintFilter-1",
+          "type": "CertificateThumbprintFilter",
+          "config": {
+            "certificate": "${pemCertificate(urlDecode(request.headers['ssl-client-cert'][0]))}",
+            "failureHandler": {
+              "type": "ScriptableHandler",
+              "config": {
+                "type": "application/x-groovy",
+                "file": "ReturnInvalidCnfKeyError.groovy"
+              }
+            }
+          }
+        },
+        {
           "comment": "Check consent authorised via AM authz policy",
           "name" : "PolicyEnforcementFilter-OBIE Assets Authorization Filter",
           "type" : "PolicyEnforcementFilter",
--- a/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
+++ b/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
@@ -42,21 +42,6 @@
           }
         },
         {
-          "comment": "Extract client certificate thumbprint for cert bound access tokens",
-          "name": "CertificateThumbprintFilter-1",
-          "type": "CertificateThumbprintFilter",
-          "config": {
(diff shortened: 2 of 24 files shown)


jorgesanchezperez (approved):

config/7.0/obdemo-bank/ig/routes/routes-service/11-ob-account-access.json
```diff
@@ -41,6 +41,21 @@
...
+          "comment": "Extract client certificate thumbprint for cert bound access tokens",
+          "name": "CertificateThumbprintFilter-1",
+          "type": "CertificateThumbprintFilter",
+          "config": {
+            "certificate": "${pemCertificate(urlDecode(request.headers['ssl-client-cert'][0]))}",
+            "failureHandler": {
+              "type": "ScriptableHandler",
+              "config": {
+                "type": "application/x-groovy",
+                "file": "ReturnTeapotError.groovy"
```
dbadham-fr, line 54: I've been meaning to ask, what is the purpose of the ReturnTeapotError.groovy? Why do we return HTTP 418?

I've never seen it used before as I thought it was a joke, I have read the RFC before but never seen it used in production code: https://www.rfc-editor.org/rfc/rfc2324#section-2.3.2

jorgesanchezperez: Good question, this is not the first time I saw this teapot error but never I have read the RFC specification, and I agree with @dbadham-fr It seems a joke.

mariantiris: No idea why teapot was returned in this case initially (most likely a joke on whoever tries to misuse a token maybe?). We can change the status to 401 since the certificate thumbprint not matching would mean you're not properly authorized.

jorgesanchezperez: I think that the status 401 will be more appropriated.

BohoCode: The [relevant specification](https://www.rfc-editor.org/rfc/rfc8705.html#section-2) states;

If no certificate is presented, or that which is presented doesn't match that which is expected for the given client_id, the authorization server returns a normal OAuth 2.0 error response per [Section 5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2) of [[RFC6749](https://www.rfc-editor.org/rfc/rfc8705.html#RFC6749)] with the invalid_client error code to indicate failed client authentication.

BohoCode (changes requested):
The [relevant specification](https://www.rfc-editor.org/rfc/rfc8705.html#section-2) states;

If no certificate is presented, or that which is presented doesn't match that which is expected for the given client_id, the authorization server returns a normal OAuth 2.0 error response per [Section 5.2](https://www.rfc-editor.org/rfc/rfc6749#section-5.2) of [[RFC6749](https://www.rfc-editor.org/rfc/rfc8705.html#RFC6749)] with the invalid_client error code to indicate failed client authentication.

jorgesanchezperez (approved):

dbadham-fr (approved):

PR #170, opened 2022-09-12
Adding failure handlers for policy validation error

mariantiris:
Issues: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/510
           https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/516

Adding a failure handler to better highlight errors return by the PolicyEnforcementFilter for the cases when the policy validation fails.

Adding fixes for functional tests, thumbprint validation, and grant validation.

### DIFF
--- a/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
+++ b/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
@@ -73,7 +73,8 @@
                   "type": "StatelessAccessTokenResolver",
                   "config": {
                     "secretsProvider": "SecretsProvider-AmJWK",
-                    "issuer": "https://&{identity.platform.fqdn}/am/oauth2/realms/root/realms/&{am.realm}"
+                    "issuer": "https://&{identity.platform.fqdn}/am/oauth2/realms/root/realms/&{am.realm}",
+                    "verificationSecretId": "any.value.in.regex.format"
                   }
                 }
               }
--- a/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
+++ b/config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
@@ -42,6 +42,21 @@
           }
         },
         {
+          "comment": "Extract client certificate thumbprint for cert bound access tokens",
+          "name": "CertificateThumbprintFilter-1",
+          "type": "CertificateThumbprintFilter",
+          "config": {
+            "certificate": "${pemCertificate(urlDecode(request.headers['ssl-client-cert'][0]))}",
+            "failureHandler": {
+              "type": "ScriptableHandler",
+              "config": {
+                "type": "application/x-groovy",
+                "file": "ReturnInvalidCnfKeyError.groovy"
+              }
+            }
+          }
+        },
+        {
           "comment": "Extracts the access_token from the request header. Uses the resolver to resolve the access_token and validate the token claims. Checks that the token has the scopes required by the filter configuration. Injects the access_token info into the OAuth2Context.",
           "name" : "OAuth2ResourceServerFilter-OB",
           "type" : "OAuth2ResourceServerFilter",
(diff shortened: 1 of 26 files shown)


config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
```diff
@@ -56,25 +71,11 @@
...
-                              "secretsProvider": "SystemAndEnvSecretStore-IAM"
-                            }
-                          }
-                        ],
-                        "handler": "ForgeRockClientHandler"
-                      }
-                    }
+                    "secretsProvider": "SecretsProvider-AmJWK",
+                    "issuer": "https://&{identity.platform.fqdn}/am/oauth2/realms/root/realms/&{am.realm}",
+                    "verificationSecretId": "any.value.in.regex.format"
```
dbadham-fr, line 78: can you explain what the verificationSecretId configuration is doing please?

config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
```diff
@@ -56,25 +71,11 @@
                 "delegate": {
                   "comment":"resolve access tokens and retrieve metadata about the token. The endpoint typically returns the time until the token expires, the OAuth 2.0 scopes associated with the token, and potentially other information",
```
dbadham-fr, line 72: can you update (or remove) the comment please, as it relates to TokenIntrospectionAccessTokenResolver

config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
```diff
@@ -56,25 +71,11 @@
...
                   "name": "token-resolver-1",
-                  "type": "TokenIntrospectionAccessTokenResolver",
+                  "type": "StatelessAccessTokenResolver",
                   "config": {
-                    "endpoint": "https://&{identity.platform.fqdn}/am/oauth2/realms/root/realms/&{am.realm}/introspect",
```
dbadham-fr, line 61: it's great to be able to achieve the functionality that we required without having to make an out of process call to AM

config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
mariantiris: The verificationSecretIdvconfigurations can be used to specify a secret ID for the secret used to verify the signature of signed access_tokens. In this case I don't think it's used (I have token over that bit from another route where that filter was already implemented in the past) because our JWT's contain the key id under the "kid" claim in the token header.

The documentation states:

The JWT signature is validated as follows:

- If the JWT contains a kid with a mapped value, for example verification.key.1.:
    - The secrets provider queries the KeyStoreSecretStore for a named secret with the secret ID verification.secret.id and the stable ID verification.key.1.
    - Because the KeyStoreSecretStore contains that mapping, the KeyStoreSecretStore returns a named secret.
    - The StatelessAccessTokenResolver tries to validate the JWT signature with the named secret. If it fails, the token is considered as invalid.


- If the JWT contains a kid with an unmapped value, for example, verification.key.3:

    - The secrets provider queries the KeyStoreSecretStore for a named secret with the secret ID verification.secret.id and the stable ID verification.key.3.
    - Because the KeyStoreSecretStore doesn’t contain that mapping, named secret resolution fails. IG tries valid secret resolution in the same way as when the JWT doesn’t contain a kid.
    - If the JWT doesn’t contain a kid:

        The secrets provider queries the KeyStoreSecretStore for all valid secrets, whose alias is mapped to the secret ID verification.secret.id

config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
mariantiris: Removed the comments since the resolver is straight forward and comment was outdated.

config/7.0/obdemo-bank/ig/routes/routes-service/10-ob-account-consent.json
mariantiris: Looks like the field is mandatory for the filter and we have to keep it

dbadham-fr (approved):

PR #171, opened 2022-09-13
497: gateway extension to convert idm intents to ob datamodel objects

jorgesanchezperez:
- Created business logic gateway conversion module
- Created filter conversion module with openig dependencies
- Created parent pom for modules
- Implementation of conversion logic
- Implementation of conversion filter
- Created unit tests coverage for all modules
- Added assembly maven plugin to build filter library with dependencies and copy it into config/7.0/obdemo-bank/ig/lib to be deployed in IG when is build
- Readme files added
- Added routes examples in {filter-module}/configuration
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/498


securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,239 @@
...
+            message.setEntity(toBytes(objectMapped));
+            if (!message.getHeaders().containsKey(ContentTypeHeader.NAME)) {
+                message.getHeaders().put(ContentTypeHeader.NAME, APPLICATION_JSON_CHARSET_UTF_8);
+            }
+            // MessageType.REQUEST
+            if (messageType.equals(MessageType.REQUEST)) {
+                return next.handle(context, (Request) message);
+            }
+            // MessageType.RESPONSE
+            return newResultPromise(((Response) message).setStatus(Status.OK));
```
dbadham-fr, line 145: What happens if a Filter after this one in the chain or the RS responds with an error such as HTTP 400?

What if an OBErrorResponse1 json payload is the message.entity? Will this be allowed to pass through or will this cause an error in the conversion logic?

jorgesanchezperez: Right! nice one! fixed!
- The filter will process the response only when the response from the called handle is successful otherwise pass through the response to the next handle.

dbadham-fr: Great, I would remove calling setStatus just to be safe

jorgesanchezperez: Set status removed.

BohoCode:
@jorgesanchezperez can you rebase this onto master - it will make it easier to review?

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,219 @@
...
+        if (logger.isInfoEnabled()) {
+            printFilterInfo();
+        }
+
+        if (messageType.equals(MessageType.REQUEST)) {
+            return process(request, request.getUri().asURI(), context, next);
+        }
+
+        return next.handle(context, request)
+                .thenAsync(response -> {
```
dbadham-fr, line 114: then can be used rather than thenAsync, as there is no async call happening in the callback. The callback applies a transformation to data which is available.

Also, you are ignoring the return value from the process method, therefore if we were to hit the catch block in that method the error response that gets created is lost.

I think it would be cleaner to extract a method which transforms the message.entity i.e. lines 130-142, then leave it up to the caller to plug the logic into the request or response flow. As the process method has become quite complex, having to deal with request and response specific concerns.

jorgesanchezperez: Improvements pushed.

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,239 @@
...
+        }
+
+        // MessageType.RESPONSE
+        return next.handle(context, request)
+                .then(response ->
+                        process(response, request.getUri().asURI())
+                )
+                .thenCatch(e ->
+                        badRequestError(e)
+                );
```
dbadham-fr, line 126: I think this can be deleted, as the process method catches Exceptions and next.handle is tagged with NeverThrowsException

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * Process the response message
+     * @param message
+     *         a {@link MessageImpl} to be processed
+     * @param uri
+     *         the request URI to extract the api version to instance the proper converter
+     * @return a {@link Promise} to be passed the next handle
+     */
```
dbadham-fr, line 136: javadoc is out of date

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
dbadham-fr: That's excellent, I think that code is much cleaner, simpler and easier to understand now.

I've added a couple of minor comments to do some cleanup, once that is done then I think it's ready to merge

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: deleted these methods to use a shared one `process` method

securebanking-openbanking-uk-gateway-conversion-filter/src/main/java/com/forgerock/securebanking/uk/gateway/conversion/filter/IntentConverterFilter.java
jorgesanchezperez: Deleted.

jorgesanchezperez:
Discard these changes to promote changes in managed IDM objects defined to make easy the serialisation/deserialisation from idm objects to ob objects and vice versa.

PR #180, opened 2022-09-27
524: Verify transport cert used during DCR 

dbadham-fr:
Verify transport cert used during DCR (dynamic client registration) is mapped to the TPP's software statement

- TLS cert is found in the HTTP header ssl-client-cert in PEM format (set by nginx)
- software_jwks_endpoint property in software statement is used to locate the TPP's JWKSet
- ProcessRegistration script now fetches the JWKSet and checks to see if a TLS cert exists by looking for a JWK with an x5c value which matches the TLS cert used to submit the DCR request
- OBIE spec notes that the x5c will only contain a single item (the client's cert) without any chain, and nginx will only send the cert on without the chain.

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/524

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
@@ -1,9 +1,12 @@
 import org.forgerock.http.protocol.*
 import org.forgerock.json.jose.*
+import org.forgerock.json.jose.jwk.*
 import org.forgerock.json.jose.common.JwtReconstruction
 import org.forgerock.json.jose.jws.SignedJwt
 import java.net.URI
-import groovy.json.JsonSlurper;
+import groovy.json.JsonSlurper
+import com.forgerock.securebanking.uk.gateway.jwks.*
+import static org.forgerock.util.promise.Promises.newResultPromise
 
 /*
  * Script to verify the registration request, and prepare AM OIDC dynamic client reg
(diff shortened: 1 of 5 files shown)


dbadham-fr:
Tweaked the impl to use nimbus to convert the PEM provided in the `ssl-client-cert` header to a JWK, so that we can compare x5c values directly. This avoids us having to manually clean the PEM format up to match the x5c format i.e. by removing the header/trailer lines and removing line breaks

dbadham-fr:
Added support for test certs generated by IG, for such certs there is no jwks_uri instead the jwks is embedded in the software statement in field `software_jwks`. In this case, the mTLS cert is validated against the embedded jwks.

Added a configuration property `security.allowIgIssuedTestCerts` to control whether the route accepts the test certs or not. Configured this property to true for the dev/config/config.json and false for prod/config/config.json

config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
```diff
@@ -167,25 +175,32 @@ switch(method.toUpperCase()) {
...
+        } else {
+            // Verify against the software_jwks which is a JWKSet embedded within the software_statement
+            // NOTE: this is only suitable for developer testing purposes
+            if (!allowIgIssuedTestCerts) {
+                return(errorResponse(Status.BAD_REQUEST, "software_statement must contain software_jwks_endpoint"));
+            }
+            logger.debug(SCRIPT_NAME + "Checking cert against ssa software_jwks: " + apiClientOrgJwks)
+            def jwkSet = new JWKSet(new JsonValue(apiClientOrgJwks.get("keys")))
+            return verifyTlsClientCertExistsInJwkSet(jwkSet)
+        }
```
BohoCode, line 203: Extract to a method???

BohoCode (approved):

config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
dbadham-fr: As discussed, it's too messy to extract a method without re-writing a lot of the filter. 

Leaving it for now, it would be nice to re-write this in Java, then we can improve the code quality + add unit tests

PR #187, opened 2022-10-10
517: Updating routes to be compatible with IDM Payments Intents schema changes

dbadham-fr:
- Updating IDM CRUD operations to work with new data model structure
- Removing ModifyPaymentIntentResponse filter as it is now redundant

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/517


config/7.1.0/securebanking/ig/scripts/groovy/ProcessPaymentConsent.groovy
```diff
@@ -20,39 +20,74 @@ def method = request.method
...
-
-
+        paymentIntentData = request.entity.getJson()
+        processProcessPaymentConsentRequestData(consentId, paymentIntentData)
+
+        def version = getObApiVersion(request)
+        def idmIntent = [
+                _id: consentId,
+                OBVersion: version,
+                OBIntentObjectType: "OBReadConsentResponse1",
```
jorgesanchezperez, line 31: `OBIntentObjectType` hardcode with the same value for all payments doesn't make sense, must be the OB object type json-based, isn't? for example `OBWriteDomesticConsentResponse5` for domestic payment v3.1.8 if we want a relevant value for `OBIntentObjectType`, if not doesn't make sense the field `OBIntentObjectType`.

config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
```diff
@@ -130,11 +116,11 @@ def convertIDMResponse(intentResponseObject, intentType) {
         case IntentType.PAYMENT_DOMESTIC_CONSENT:
             responseObj = [
                     "id"                   : intentResponseObject._id,
-                    "data"                 : intentResponseObject.Data,
+                    "data"                 : intentResponseObject.OBIntentObject.Data,
                     // RS expect 'Data' instead 'data' to deserialize the json to OB object
```
jorgesanchezperez, line 120: I suspect that keep the duplication of `d/Data` is to keep the compatibility until finished the epic task, right?

jorgesanchezperez (approved):

config/7.1.0/securebanking/ig/scripts/groovy/ProcessPaymentConsent.groovy
dbadham-fr: Ooops, good spot!

config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
dbadham-fr: Yes, I will review the RCS code and clean this up afterwards. If we can get this set of changes merged then the next payment APIs can build on top of it, this cleanup can then be done without impacting anyone I think

config/7.1.0/securebanking/ig/scripts/groovy/ProcessPaymentConsent.groovy
dbadham-fr: To fix we will add a route argument to enable the OBIntentObjectType to be configured per payment route.

For now, we will hardcode this as the response type supported by OBIE Read/Write specs 3.1.8-3.1.10, this is a temporary solution until we move the processing of consents to the RS. 

Within the RS it will be much simpler to determine the OBIntentObjectType (and OBVersion) values, as these will be known to each ConsentApiController implementation.

jorgesanchezperez (approved):

PR #193, opened 2022-10-12
573: calculated elements file payments

jorgesanchezperez:
- Deleted specific cases to convertIDM responses for payments to use the default case and support all payment intent types
Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/573

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
@@ -109,7 +109,7 @@ def convertIDMResponse(intentResponseObject, intentType) {
                     "oauth2ClientName"     : intentResponseObject.apiClient.name
             ]
             break
-        else: // For all payments intents
+        default: // For all payments intents
             responseObj = [
                     "id"                   : intentResponseObject._id,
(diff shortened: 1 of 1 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
```diff
@@ -109,61 +109,14 @@ def convertIDMResponse(intentResponseObject, intentType) {
                     "oauth2ClientName"     : intentResponseObject.apiClient.name
             ]
             break
-        case IntentType.PAYMENT_DOMESTIC_CONSENT:
+        else: // For all payments intents
```
dbadham-fr, line 112: Please test this, the syntax looks wrong to me as you still have a switch statement but now there is an else block?

dbadham-fr (changes requested):

config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
jorgesanchezperez: I don't know in what I was thinking when I added an `else` as case instead of `default`... Fixed.

dbadham-fr (approved):

PR #202, opened 2022-10-27
615: Fix FAPI test CallAccountRequestsEndpointWithBearerToken

BohoCode:
The following FAPI test is failing: CallAccountRequestsEndpointWithBearerToken This is because the FAPI conformance is providing a jwt with an obdemo.<env> issuer, and the route was checking for an iam.<env>.doman issuer

Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/615


dbadham-fr:
Closing, work was done on a FAPI branch which was then merged.

PR #203, opened 2022-10-28
597: Fixing bug where idm apiClient is being deleted even if the AM delete is rejected

dbadham-fr:
CreateApiClient filter runs after the response from the ReverseProxyHandler that talks to AM. The apiClient must only be deleted if the AM response is successful i.e. AM delete the OAuth2 client

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/597


dbadham-fr:
Closing PR, rebase has included some unexpected diffs, will reapply changes to a new branch and raise a new PR

PR #212, opened 2022-11-02
621: Only add software_statement to successful AM responses

dbadham-fr:
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/621

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
@@ -279,11 +279,13 @@ private void rewriteUriToAccessExistingAmRegistration() {
 }
 
 private void addSoftwareStatementToResponse(response, ssa) {
-    var registrationResponse = response.getEntity().getJson()
-    if (!registrationResponse["software_statement"]) {
-        registrationResponse["software_statement"] = ssa
+    if (response.status.isSuccessful()) {
+        var registrationResponse = response.getEntity().getJson()
+        if (!registrationResponse["software_statement"]) {
+            registrationResponse["software_statement"] = ssa
+        }
+        response.entity.setJson(registrationResponse)
     }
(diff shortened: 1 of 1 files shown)


jorgesanchezperez (approved):

config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
```diff
@@ -279,11 +279,13 @@ private void rewriteUriToAccessExistingAmRegistration() {
...
 private void addSoftwareStatementToResponse(response, ssa) {
-    var registrationResponse = response.getEntity().getJson()
-    if (!registrationResponse["software_statement"]) {
-        registrationResponse["software_statement"] = ssa
+    if (response.status.isSuccessful()) {
+        var registrationResponse = response.getEntity().getJson()
+        if (!registrationResponse["software_statement"]) {
+            registrationResponse["software_statement"] = ssa
+        }
+        response.entity.setJson(registrationResponse)
```
BohoCode, line 287: Do we make a good identifiable log entry on failure (all types of failure) that customers can use when TPPs contact them to say "I can't register"?

BohoCode (approved):

config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
dbadham-fr: No, there is no general error handler.

If validation in ProcessRegistration.groovy fails, then a message is logged and a HTTP Error response returned.

If the failure is downstream in AM, then the AM response is just bubbled up through the filters (and nothing gets logged other than the raw AM response).

I've raised an issue to review this (as I think it is outside of the scope of 621): https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/641

PR #217, opened 2022-11-04
Update AddDetachedSig.groovy

mariantiris:
Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/660

Added an additional check for empty payload on response signing. An empty JwtClaimsSet object is initialised in the case of empty response payload, which allows us to create empty reponse signature as required by the spec in the case of the submit file API **POST /file-payment-consents/{ConsentId}/file**

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
@@ -36,7 +36,13 @@ next.handle(context, request).thenOnResult({ response ->
     signingManager.newSigningHandler(purpose).then({ signingHandler ->
         logger.debug(SCRIPT_NAME + "Building of the JWT started")
 
-        JwtClaimsSet jwtClaimsSet = new JwtClaimsSet(response.getEntity().getJson())
+        JwtClaimsSet jwtClaimsSet
+        // We get content empty on submit file payment API
+        if (response.getEntity().isRawContentEmpty()) {
+            jwtClaimsSet = new JwtClaimsSet()
+        } else {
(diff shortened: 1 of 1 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
```diff
@@ -36,7 +36,13 @@ next.handle(context, request).thenOnResult({ response ->
     signingManager.newSigningHandler(purpose).then({ signingHandler ->
         logger.debug(SCRIPT_NAME + "Building of the JWT started")
 
-        JwtClaimsSet jwtClaimsSet = new JwtClaimsSet(response.getEntity().getJson())
+        JwtClaimsSet jwtClaimsSet
+        // We get content empty on submit file payment API
+        if (response.getEntity().isRawContentEmpty()) {
+            jwtClaimsSet = new JwtClaimsSet()
+        } else {
```
jorgesanchezperez, line 43: When the content is empty we need to return an error, doesn't make sense process the request because never will match with the initiation object send in the consent previously.

jorgesanchezperez (changes requested):
response with error when upload a file with empty content.

config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
mariantiris: It's not an error case. It'a a success case where the response payload is empty. We need to create am empty signature from what I understand from the spec (on the response of submit file payment file the response on the API is 200 OK with empty body). This is the case that was failing.

jorgesanchezperez (approved):

PR #218, opened 2022-11-07
257: Propagate x-fapi-interaction-id through the system

BohoCode:
Create a filter as suggested by Guillaume Sauthier in the following issue: https://bugster.forgerock.org/jira/browse/OPENIG-6538

The aim is to set the transactionIdContext to be the value of the x-fapi-interaction-id. This _should_ ensure that the value of the x-fapi-interaction-id assigned by IG is passed to AM & IDM as the x-forgerock-transactionid header and will make logs traceable across the system.

Also in this PR:
1. Modify routes to set the x-fapi-interaction-id if it is not set
2. Set the transaction context to the value of the x-fapi-interaction-id. This results in the x-forgerock-transactionid passed to requests to the FR Platform to have the same value as the x-fapi-interaction-id
3. Modify the IG scripts to include the x-fapi-interaction-id in their log messages

The route all now have the "SBATFapiInteractionFilterChain" defined at the start of the filter chain. This is a ChainOfFilters that performs actions 1 & 2 above. 

Observations: IDM is not using the x-forgerock-transactionid for it's logging. Further work will be needed to make this happen.

Issue: https://github.com/securebankingaccesstoolkit/securebankingaccesstoolkit/issues/257

### DIFF
--- a/config/7.1.0/securebanking/ig/config/dev/config/config.json
+++ b/config/7.1.0/securebanking/ig/config/dev/config/config.json
@@ -69,11 +69,55 @@
       "name" : "ForgeRockClientHandler",
       "type" : "Chain",
       "config" : {
-        "filters" : [ "TransactionIdOutboundFilter" ],
+        "filters" : [ 
+          {
+            "comment": "Add x-fapi-interaction-id header if one was not present in the request",
+            "name": "FapiInteractionIdFilter",
+            "type": "FapiInteractionIdFilter"
+          },
+          {
+            "comment": "Copy the x-fapi-interaction-id header to TransactionIdContext",
+            "name": "FapiTransactionIdFilter-1",
+            "type": "ScriptableFilter",
+            "config": {
+              "type": "application/x-groovy",
+              "file": "FapiTransactionIdFilter.groovy"
+            }
+          },
+          "TransactionIdOutboundFilter"
+       ],
         "handler" : "ClientHandler"
       },
       "capture" : [ "response", "request" ]
     },
+    {
+      "name": "SBATReverseProxyHandler",
+      "type": "Chain",
+      "capture": [
+        "request",
+        "response"
+      ],
+      "config": {
+        "filters" : [ 
+          {
+            "comment": "Add x-fapi-interaction-id header if one was not present in the request",
+            "name": "FapiInteractionIdFilter",
+            "type": "FapiInteractionIdFilter"
+          },
+          {
+            "comment": "Copy the x-fapi-interaction-id header to TransactionIdContext",
+            "name": "FapiTransactionIdFilter-1",
+            "type": "ScriptableFilter",
+            "config": {
+              "type": "application/x-groovy",
+              "file": "FapiTransactionIdFilter.groovy"
+            }
+          },
+          "TransactionIdOutboundFilter"
+       ],
+        "handler" : "ReverseProxyHandler"
+      }
+    },    
     {
       "name" : "AmService-OBIE",
       "type" : "AmService",
--- a/config/7.1.0/securebanking/ig/config/dev/config/config.json
+++ b/config/7.1.0/securebanking/ig/config/dev/config/config.json
@@ -79,6 +79,7 @@
     {
       "name": "SBATFapiInteractionFilterChain",
       "type": "ChainOfFilters",
+      "comment": "This filter chain will set the x-fapi-interaction-id (if not provided in the request), and also set the transaction context to the x-fapi-interaction-id value. This means that if the 'TransactionIdOutboundFilter' is specified on any handlers used by the chain the x-fapi-interaction-id value will be passed onward in the X-ForgeRock-TransactionId header",
       "config" : {
         "filters": [
           {
--- a/config/7.1.0/securebanking/ig/config/dev/config/config.json
+++ b/config/7.1.0/securebanking/ig/config/dev/config/config.json
@@ -70,6 +70,17 @@
       "type" : "Chain",
       "config" : {
         "filters" : [ 
+          "TransactionIdOutboundFilter"
+       ],
+        "handler" : "ClientHandler"
+      },
+      "capture" : [ "response", "request" ]
+    },
+    {
+      "name": "SBATFapiInteractionFilterChain",
+      "type": "ChainOfFilters",
+      "config" : {
+        "filters": [
           {
             "comment": "Add x-fapi-interaction-id header if one was not present in the request",
             "name": "FapiInteractionIdFilter",
@@ -83,12 +94,9 @@
               "type": "application/x-groovy",
               "file": "FapiTransactionIdFilter.groovy"
             }
-          },
-          "TransactionIdOutboundFilter"
-       ],
-        "handler" : "ClientHandler"
-      },
-      "capture" : [ "response", "request" ]
+          }
+        ]
+      }
     },
     {
       "name": "SBATReverseProxyHandler",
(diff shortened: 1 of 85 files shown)


jorgesanchezperez (dismissed):
Seems right to me!

config/7.1.0/securebanking/ig/config/dev/config/config.json
```diff
@@ -69,18 +69,62 @@
...
+          },
+          {
+            "comment": "Copy the x-fapi-interaction-id header to TransactionIdContext",
+            "name": "FapiTransactionIdFilter-1",
+            "type": "ScriptableFilter",
+            "config": {
+              "type": "application/x-groovy",
+              "file": "FapiTransactionIdFilter.groovy"
+            }
+          },
```
dbadham-fr, line 86: I think each route's filter chain should start with these 2 filters. These filters should then be removed from the `ForgeRockClientHandler` and `SBATReverseProxyHandler`.

It makes sense to add the `TransactionIdOutboundFilter` to the `SBATReverseProxyHandler` (which you have done). I think the `ForgeRockClientHandler` should be reverted to it's previous state.

Both of the filters need to be at the start of each route's chain in order to get the transactionId into the context as earlier as possible, this will then be passed down to other FR products either by the FRClientHandler or the RPHandler. Adding them to the handler's themselves is too late, other filters will have already run and potentially need ed to log the value.

It is unfortunate to have to duplicate the config across all routes, but it seems the best way to go to me.

config/7.1.0/securebanking/ig/config/dev/config/config.json
```diff
@@ -69,18 +69,62 @@
...
+      }
+    },    
     {
       "name" : "AmService-OBIE",
       "type" : "AmService",
       "config" : {
         "url" : "https://&{identity.platform.fqdn}/am",
         "realm" : "/&{am.realm}",
-        "version" : "6.5.1",
+        "version" : "7.2.0",
```
dbadham-fr, line 127: Nice spot. didn't know about this, hopefully it magically fixes my AM redirect issue...

config/7.1.0/securebanking/ig/config/dev/config/config.json
BohoCode: Yeah, I think you are right Dan. Seems a waste to have to duplicate. I could create a chain with both in and then only one definition in each route.

config/7.1.0/securebanking/ig/config/dev/config/config.json
BohoCode: I don't think it will. I believe IG tried to determine the version automagically and if it can find it out it will use that in preference. I think we're no using v 7.2.0 of IG though, which might cause an issue!!!

config/7.1.0/securebanking/ig/config/dev/config/config.json
BohoCode: Ah! https://backstage.forgerock.com/docs/ig/7.2/reference/ChainOfFilters.html

config/7.1.0/securebanking/ig/config/dev/config/config.json
```diff
@@ -69,18 +69,56 @@
...
+        "filters" : [ 
+          "TransactionIdOutboundFilter"
+       ],
         "handler" : "ClientHandler"
       },
       "capture" : [ "response", "request" ]
     },
+    {
+      "name": "SBATFapiInteractionFilterChain",
+      "type": "ChainOfFilters",
```
dbadham-fr, line 81: Nice

config/7.1.0/securebanking/ig/routes/routes-service/01-ob-rs-metadata.json
```diff
@@ -56,11 +57,11 @@
           "type": "ScriptableFilter",
           "config": {
             "type": "application/x-groovy",
-            "file" : "ProcessRs.groovy"
+            "file": "ProcessRs.groovy"
           }
         }
       ],
-      "handler": "ReverseProxyHandler"
+      "handler": "ForgeRockClientHandler"
```
dbadham-fr, line 64: Why change this?
IG is reverse proxying the RS for this route

docker/7.1.0/ig/Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM gcr.io/forgerock-io/ig:7.1.0
 # Copy all config files into the docker image.
 # The default ig directory is /var/ig, and it expects subfolders config/ and scripts/ (if required)
 
+ENV JAVA_OPTS="-Dorg.forgerock.http.TrustTransactionHeader=true"
```
dbadham-fr, line 6: Is this flag needed for IG?

I believe the flag allows IG to trust an `x-forgerock-transaction-id` header passed in, but for our use case the TPP App should never be supplying this, instead they should send the `x-fapi-interaction-id`

Other FR apps won't response with this header, and IG is the entrypoint into other FR apps, therefore I don't see the use case.

dbadham-fr (changes requested):

config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
```diff
@@ -16,7 +16,10 @@ import org.forgerock.http.protocol.Status
  *
  */
 
-SCRIPT_NAME = "[AddDetachedSig] - "
+def fapiInteractionId = request.getHeaders().getFirst("x-fapi-interaction-id");
+if(fapiInteractionId == null) fapiInteractionId = "No x-fapi-interaction-id"
+SCRIPT_NAME = "[AddDetachedSig] (" + fapiInteractionId + ") - "
```
mariantiris, line 21: I wonder if it would be good to add something like `%h x-fapi-interaction-id: %header{x-fapi-interaction-id}` in the logback.xml pattern for all lines logged in IG.

Would look something like: <pattern>%d{HH:mm:ss:SSS} | %-5level | %thread | %logger{20} | %h x-fapi-interaction-id: %header{x-fapi-interaction-id}|%message%n%xException</pattern>

BohoCode: Looks like a good suggestion!

config/7.1.0/securebanking/ig/routes/routes-service/01-ob-rs-metadata.json
BohoCode: Ah good catch!

BohoCode: Fixed

docker/7.1.0/ig/Dockerfile
BohoCode: Fixed

config/7.1.0/securebanking/ig/scripts/groovy/AddDetachedSig.groovy
BohoCode: Thinking about this, we are dependent on customers then configuring logback correctly, and I think we might end up using the forgeops project to deploy IG, and just use our repo to build the mounted IG config (routes, scripts, jars) etc. So this would end up being customer config dependent, or require modifying the base configuration of IG and building their own IG image.

So I wonder if this method is more reliable and is one thing less for customers to have to get right?

BohoCode: Tried this, but didn't get far with it. Will create a new issue to look at this at a later date.

dbadham-fr (approved):

PR #220, opened 2022-11-09
624: Validating the DCR JWT signature

dbadham-fr:
Adding RsaJwtSignatureValidator which validates that a SignedJwt's signature can be verified against a signing key in a supplied JWKSet

Calling the new validator from ProcessRegistration filter to verify the signature of the DCR JWT. 

Refactoring the SSAVerifier filter to use the new validator, instead of its validation written in groovy, to validate the signature of the SSA.

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/624


BohoCode (approved):
Nice!

PR #225, opened 2022-11-15
673: Removing FR private repos from pom

dbadham-fr:


dbadham-fr:
Private releases is required for the openig dependency.

PR #226, opened 2022-11-15
673: Configuring x-ob-url header in IG

dbadham-fr:
Configuring the x-ob-url header required by the RS backends for account APIs.

Setting the value to match the IG route url as accessed by the TPP.

NOTE: contexts.router.originalUri could also be used, but it would need editing to replace the http:// with https:// (as nginx is terminating TLS in front of IG)


BohoCode (commented):
Could add this filter to the SBATFapiInteractionFilterChain in config.json so it applies to all routes? Would need to rebase this branch onto master for that though?

dbadham-fr:
Closing PR, to apply the changes as suggested by Jamie it is quicker to create a new branch.

PR #228, opened 2022-11-15
673: Creating separate ReverseProxyHandlers for calls to RS vs Identity Platform

dbadham-fr:
ReverseProxyHandlers:
- SBATReverseProxyHandlerIdentityPlatform (calls to AM or IDM)
- SBATReverseProxyHandlerRs

Only adding the x-ob-url header when reverse proxying the RS


BohoCode (approved):
Ah, that's great to separate them out.

PR #234, opened 2022-11-18
641: Improve DCR error handling

dbadham-fr:
Improving DCR error handling by creating an error factory which can be used to consistently report DCR errors. The error format for bad client requests adheres to the spec: https://www.rfc-editor.org/rfc/rfc7591#section-3.2.2

[ErrorResponseFactory](https://github.com/SecureBankingAccessToolkit/securebanking-openbanking-uk-gateway/blob/6b5b3f7e1c138d22463736d5b0850432362b98b4/config/7.1.0/securebanking/ig/scripts/groovy/com/securebanking/gateway/dcr/ErrorResponseFactory.groovy) is a groovy class which is imported into the ProcessRegistration and SSAVerifier filter scripts which make up the core DCR logic. 

This pattern can be applied to reuse logic between groovy filters. To do this place groovy classes under scripts/groovy dir in a standard groovy package structure. You can then import the class in scripts as you would any other groovy/java class  in this case: `import com.securebanking.gateway.dcr.ErrorResponseFactory`

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/641

### DIFF
--- a/config/7.1.0/securebanking/ig/routes/routes-service/03-ob-dcr-register-tpp.json
+++ b/config/7.1.0/securebanking/ig/routes/routes-service/03-ob-dcr-register-tpp.json
@@ -67,8 +67,8 @@
             "clientHandler": "OBClientHandler",
             "args": {
               "routeArgSSAIssuerJwksUrls": {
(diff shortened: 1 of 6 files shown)


jorgesanchezperez (approved):
Nice

PR #239, opened 2022-11-23
655: Validating DCR redirect_uris

dbadham-fr:
The redirect_uris must match or be a subset of the software_statement.software_redirect_uris

If no redirect_uris are present in the registration request, then all the software_statement.software_redirect_uris will be used

The ob spec also requires that redirect_uris use https and do not have localhost as the host

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/655


config/7.1.0/securebanking/ig/scripts/groovy/ProcessRegistration.groovy
```diff
@@ -302,3 +307,38 @@ private boolean validateRegistrationJwtSignature(jwt, jwkSet) {
         return false
     }
 }
+
```
BohoCode, line 310: Method comment with link to relevant specs might be useful for those following....

dbadham-fr: Done

BohoCode (approved):

PR #240, opened 2022-11-24
683: Refactoring CreateApiClient to use Promise API correctly

dbadham-fr:
Changes to CreateApiClient POST handling:
- Properly chained Promises using thenAsync when making http calls in response handling i.e. calling IDM as part of handling the AM response
- Fixed error handling
  - Return Response objects or Promises wrapping Response objects as appropriate
  - Unexpected errors from IDM will bubble up as HTTP 500 errors
- Changed order of calls to create the apiClientOrg first (if it does not already exist), then creating the apiClient
  - Using a PUT with header "If-None-Match: *" to prevent existing apiClientOrg objects being overwritten

Useful CREST documentation:
https://backstage.forgerock.com/docs/idm/7.2/crest/crest-create.html
https://backstage.forgerock.com/docs/idm/7.2/crest/crest-status-codes.html

Issue: https://github.com/SecureBankingAccessToolkit/SecureBankingAccessToolkit/issues/683

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/CreateApiClient.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/CreateApiClient.groovy
@@ -4,9 +4,11 @@ import org.forgerock.json.jose.*
 import static org.forgerock.util.promise.Promises.newResultPromise
 
 /*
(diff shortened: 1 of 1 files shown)


mariantiris (approved):

jorgesanchezperez (approved):
Looks good to me.

PR #258, opened 2022-12-22
729: FetchApiClientFilter impl which queries IDM for ApiClient data

dbadham-fr:
Added data model classes ApiClient and ApiClientOrganistation.

FetchApiClientFilter queries IDM for ApiClient data based on the client_id found in the OAuth2Context. The filter than adds the ApiClient as a context attribute so that it can be used by other filters. The groovy filters this can be accessed via `attributes.apiClient`

NOTE: As the filter depends on the OAuth2Context then it must be installed in the chain after a filter which adds this context, typically: OAuth2ResourceServerFilter.

Added the filter to payment routes and used it in ProcessDetachedSig script to simplify accessing the ApiClient data.

In future the filter will be added to more routes, and other filters within the payments route will use the data.

### DIFF
--- a/.github/workflows/slackNotification.yml
+++ b/.github/workflows/slackNotification.yml
@@ -1,0 +1,18 @@
+name: Check Changes
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    name: Push Slack Notification
+    steps:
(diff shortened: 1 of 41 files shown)


dbadham-fr:
Closed in favour of:
https://github.com/SecureApiGateway/securebanking-openbanking-uk-gateway/pull/261

PR #260, opened 2022-12-28
720 - Slack notification when PR complete

shaunharrisonFR:
New github action workflow which will trigger when there is a push to master, push to master is only possible when a PR is complete so does the same result as a merge being closed

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/720

### DIFF
--- a/.github/workflows/slackNotification.yml
+++ b/.github/workflows/slackNotification.yml
@@ -1,0 +1,18 @@
+name: Check Changes
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    name: Push Slack Notification
+    steps:
+      - name: Slack Notification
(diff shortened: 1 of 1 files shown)


dbadham-fr:
Can you share an example of the slack message please?

shaunharrisonFR:
image

This is one from when I was doing the work, the name of the bot should now be github actions and have a different icon

dbadham-fr:
Nice, so all the formatting etc comes from the action: `https://github.com/rtCamp/action-slack-notify` ?

dbadham-fr (approved):

shaunharrisonFR:
Yeah that's right, there are some options we  can overwrite but kept it simple for now

dbadham-fr:
Looks good, let's merge this one and see how it looks then I'll approve the rest

PR #262, opened 2023-01-03
689: TransportCertValidationFilter

dbadham-fr:
Adding TransportCertValidationFilter which validates that the client's MTLS certificate is valid by checking that it exists in their JWKSet (as determined by their ApiClient and TrustedDirectory attributes).

The validation logic is applied by a TransportCertValidator instance, adding DefaultTransportCertValidator implementation which checks for a certificate match by transforming the incoming cert into a JWK and comparing the JWK.x5c value against the values in the JWKSet for the ApiClient.

Adding the filter to FetchApiClientResourcesChain so that the validation is applied to all routes.

https://github.com/SecureApiGateway/SecureApiGateway/issues/689

### DIFF
--- a/.github/workflows/slackNotification.yml
+++ b/.github/workflows/slackNotification.yml
@@ -1,0 +1,18 @@
+name: Check Changes
+
+on:
+  push:
+    branches:
+      - master
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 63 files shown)


BohoCode (approved):
Very clean. We need to start working out how we document all these filters we're creating!

PR #264, opened 2023-01-04
729: Adding FetchTrustedDirectoryFilter

dbadham-fr:
This filter will add the TrustedDirectory for this ApiClient to the attributes context. NOTE: FetchTrustedDirectoryFilter should be installed in a filter chain after the FetchApiClientFilter filter, as the ApiClient data is required to determine the TrustedDirectory.

Added filter chain: FetchApiClientAndTrustedDirectoryDataChain which can be used to fetch the ApiClient and then the TrustedDirectory. Updating existing usages of the FetchApiClientFilter in routes to use the new chain.

### DIFF
--- a/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
+++ b/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
@@ -44,6 +44,9 @@ import com.forgerock.sapi.gateway.fapi.FAPIUtils;
  */
 public class FetchTrustedDirectoryFilter implements Filter {
 
+    /**
+     * The key to use to get the TrustedDirectory from the AttributesContext
+     */
     public static final String TRUSTED_DIRECTORY_ATTR_KEY = "trustedDirectory";
 
     private final Logger logger = LoggerFactory.getLogger(getClass());
--- a/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
+++ b/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
@@ -1,0 +1,124 @@
+/*
+ * Copyright © 2020-2022 ForgeRock AS (obst@forgerock.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.trusteddirectories;
+
+import static org.forgerock.openig.util.JsonValues.requiredHeapObject;
+
+import org.forgerock.http.Filter;
+import org.forgerock.http.Handler;
+import org.forgerock.http.protocol.Request;
+import org.forgerock.http.protocol.Response;
+import org.forgerock.json.jose.jwt.JwtClaimsSet;
+import org.forgerock.openig.heap.GenericHeaplet;
+import org.forgerock.openig.heap.HeapException;
+import org.forgerock.services.context.AttributesContext;
+import org.forgerock.services.context.Context;
+import org.forgerock.util.Reject;
+import org.forgerock.util.promise.NeverThrowsException;
+import org.forgerock.util.promise.Promise;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.forgerock.sapi.gateway.dcr.ApiClient;
+import com.forgerock.sapi.gateway.dcr.FetchApiClientFilter;
+import com.forgerock.sapi.gateway.fapi.FAPIUtils;
+
+/**
+ * Fetches {@link TrustedDirectory} configuration for the {@link ApiClient} that is configured in the {@link AttributesContext}
+ * and adds the TrustedDirectory to the attributes context so that it can be used by subsequent filters.
+ *
+ * This filter must be installed after a filter which adds the ApiClient to the context, typically: {@link FetchApiClientFilter}
+ */
+public class FetchTrustedDirectoryFilter implements Filter {
+
+    public static final String TRUSTED_DIRECTORY_ATTR_KEY = "trustedDirectory";
+
+    private final Logger logger = LoggerFactory.getLogger(getClass());
+
+    /**
+     * The Service used to retrieve the TrustedDirectory
+     */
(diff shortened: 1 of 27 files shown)


secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
```diff
@@ -0,0 +1,127 @@
...
+    /**
+     * Utility method to retrieve a TrustedDirectory object from a Context.
+     * This method can be used by other filters to retrieve the TrustedDirectory installed into the attributes context by
+     * this filter.
+     *
+     * @param context the context to retrieve the TrustedDirectory from
+     * @return the TrustedDirectory or null if it is not set in the context.
+     */
+    public static TrustedDirectory getTrustedDirectoryFromContext(Context context) {
+        return (TrustedDirectory) context.asContext(AttributesContext.class).getAttributes().get(TRUSTED_DIRECTORY_ATTR_KEY);
```
BohoCode, line 68: That line looks fragile and hard to work out what went wrong it if doesn't work. Maybe it could be split and some error handling added? At least some logs if for example there are no attributes rather than an npe?

BohoCode: If customers are using these filters then we need to make error handling/logging as helpful as poss I think.

BohoCode: asContext can throw an unchecked IllegalArgumentException. We should use checked exceptions for filter APIs? What is IG best practice on this? Do they try to prevent unchecked exceptions being thrown from filters etc?

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
```diff
@@ -0,0 +1,127 @@
...
+            context.asContext(AttributesContext.class).getAttributes().put(TRUSTED_DIRECTORY_ATTR_KEY, getTrustedDirectory(apiClient));
+            return next.handle(context, request);
+        } catch (RuntimeException ex) {
+            logger.error("(" + FAPIUtils.getFapiInteractionIdForDisplay(context) + ") failed to get trustedDirectory for apiClient: " + apiClient, ex);
+            throw ex;
+        }
+    }
+
+    private TrustedDirectory getTrustedDirectory(ApiClient apiClient) {
+        final JwtClaimsSet ssaClaims = apiClient.getSoftwareStatementAssertion().getClaimsSet();
```
BohoCode, line 93: What if apiClient doesn't have a software statement assertion for some reason? Maybe someone nulled it down in IDM?

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
```diff
@@ -0,0 +1,127 @@
...
+        return (TrustedDirectory) context.asContext(AttributesContext.class).getAttributes().get(TRUSTED_DIRECTORY_ATTR_KEY);
+    }
+
+    public FetchTrustedDirectoryFilter(TrustedDirectoryService trustedDirectoryService) {
+        Reject.ifNull(trustedDirectoryService, "trustedDirectoryService must be provided");
+        this.trustedDirectoryService = trustedDirectoryService;
+    }
+
+    @Override
+    public Promise<Response, NeverThrowsException> filter(Context context, Request request, Handler next) {
```
BohoCode, line 77: Ah, what's that NeverThrowsException do?? Maybe it answers all my questions? Still some logging and nice error messages still nice to have

secure-api-gateway-ig-extensions/src/test/java/com/forgerock/sapi/gateway/util/TestHandlers.java
```diff
@@ -0,0 +1,71 @@
...
+import org.forgerock.http.protocol.Response;
+import org.forgerock.http.protocol.Status;
+import org.forgerock.services.context.Context;
+import org.forgerock.util.promise.NeverThrowsException;
+import org.forgerock.util.promise.Promise;
+import org.forgerock.util.promise.Promises;
+
+/**
+ * Collection of {@link Handler} implementations which are useful for testing purposes
+ */
```
BohoCode, line 31: That's useful!

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
dbadham-fr: The line won't NPE, a null will be returned if the attribute isn't set. This particular behaviour is being tested here: https://github.com/SecureApiGateway/securebanking-openbanking-uk-gateway/pull/264/files#diff-5efadbf661ea274fee482bfc776aea620f4af47580cd6b85adaab5ae2d4fa3caR115-R116 , we call the getTrustedDirectoryFromContext at the start of the test to verify that there is no attribute prior to calling the filter.

The Context can throw IllegalArgumentException if it does not contain AttributesContext.
However, AttributesContext is added into the Context hierarchy as standard in IG and is available to all Handlers in the chains that we define.

My opinion is that if the AttributesContext is missing then it is a programming error and a RuntimeException is acceptable. IG will then trap this and return a 500 error.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
dbadham-fr: ssa is a required field in IDM, the FetchApiClientFilter will fail to decode the response if the field is missing and we will not reach this filter.

I'll add a unit test for this case to the FetchApiClientFilterTest

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
dbadham-fr: It is a marker to show that a methods never throws an exception. 

All of the HTTP based methods will have this sig, it encourages you to return a HTTP based error message.

Runtime exceptions may still be thrown, the `org.forgerock.openig.filter.RuntimeExceptionFilter` acts as an uncaught exception handler in the chain and will catch RuntimeExceptions and return a HTTP 500 response (if they are not caught and handled by another filter in the chain).

BohoCode (dismissed):

BohoCode (approved):

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/FetchTrustedDirectoryFilter.java
BohoCode: Nice!

PR #265, opened 2023-01-04
773: process registration should not rely on certificate

BohoCode:
Based on the description of how Dynamic Client Registration should work [here](https://github.com/SecureApiGateway/SecureApiGateway/wiki/About-Dynamic-Client-Registration) we need to ensure that ProcessRegistration.groovy:

    Does not depend upon details within the certificate. This is brittle - we should not require the presented cert to be in eIDAS form.
    Relies upon using the TrustedDirectoriesService to determine the directory from which a software_statement was issued and use the trusted directory JWKS or jwks_uri to validate the certificate rather than trying to inspect the certificate.
    Performs the necessary checks defined in the OB Dynamic Client Registration specification

Issue: https://github.com/secureapigateway/secureapigateway/issues/733

### DIFF
--- a/config/7.1.0/securebanking/ig/config/dev/config/config.json
+++ b/config/7.1.0/securebanking/ig/config/dev/config/config.json
@@ -395,6 +395,7 @@
       "name": "OBJwkSetService",
       "type": "CaffeineCachingJwkSetService",
       "config": {
+        "handler": "OBClientHandler",
         "maxCacheEntries": 500,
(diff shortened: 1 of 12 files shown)


BohoCode:
**NOTE:** This work hasn't really finished. We need to review the route in general and the order in which we do things. Need to do registrationJwt and SSA validation up front, then do OB related checks, then pass off to AM to do OAuth2 checks..

dbadham-fr (approved):

PR #267, opened 2023-01-05
729: Adding FetchApiClientJwksFilter

dbadham-fr:
FetchApiClientJwksFilter uses the ApiClient and TrustedDirectory objects from the context to lookup and set the apiClientJwkSet attribute. This attribute contains the JWKSet object which holds the keys that the ApiClient has registered with the TrustedDirectory.

Updated ProcessDetachedSig filter to use the apiClientJwkSet attribute to find the key to use when validating the signature.


BohoCode (approved):
Fantastic! :-D

PR #268, opened 2023-01-06
733: Improve and rework the 03-dcr-register-tpp route

BohoCode:
The order in which we do things in the 03-dcr-register-tpp route doesn't make too much sense. We should do checks in this order:

    Validate the SSA & registrationJwt
    Perform FAPI validation
    Perform OB specific checks
    Hand off to AM to do OAuth2 checks

We should use the TrustedDirectoriesService were applicable and not have hard coded references to claims within the SSA jwt. We should get the claim names for values we need from the trusted directories.

We should not depend upon the format of the presented certificates, i.e. we should not assume that they are of an eIDAS form. We should depend upon the Trusted Directory to have done all the checks on the certificate, and as long as the certificate can be validated against the Software Statement JWKS then we can trust the certificate and the roles in the SSA.

For further info on how Dynamic Client Registration should work see [here](https://github.com/SecureApiGateway/SecureApiGateway/wiki/About-Dynamic-Client-Registration)

Issue: https://github.com/secureapigateway/secureapigateway/issues/733


BohoCode:
This PR contains the new Filter. 

Still need to review the route as a whole

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSsaSignatureValidator.java
```diff
@@ -0,0 +1,155 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.dcr.request;
+
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.security.SignatureException;
+
+import javax.validation.constraints.NotNull;
```
dbadham-fr, line 22: Why is the Java Bean Validation API being used? 

This API is typically aimed at validating user input, it is used in the Open Banking datamodel as part of the generated code. Spring can be configured to process these annotations and apply the validation.

IG will not be processing these annotations, so the validation will not be being applied.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSignatureValidationRuntimeException.java
```diff
@@ -0,0 +1,27 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.dcr.request;
+
+public class DCRSignatureValidationRuntimeException extends RuntimeException{
```
dbadham-fr, line 18: what is the benefit of this over throwing a RuntimeException which comes with the jvm such as IllegalArgumentException or IllegalStateException?

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
```diff
@@ -0,0 +1,319 @@
...
+
+import com.forgerock.sapi.gateway.dcr.request.DCRSignatureValidationException.ErrorCode;
+import com.forgerock.sapi.gateway.dcr.utils.DCRUtils;
+import com.forgerock.sapi.gateway.dcr.utils.JwtReconstructionException;
+import com.forgerock.sapi.gateway.fapi.FAPIUtils;
+import com.forgerock.sapi.gateway.jwks.JwkSetService;
+import com.forgerock.sapi.gateway.jws.JwtSignatureValidator;
+import com.forgerock.sapi.gateway.trusteddirectories.TrustedDirectoryService;
+
+public class RequestAndSsaSignatureValidationFilter implements Filter {
```
dbadham-fr, line 60: I think this should be renamed, the current name suggests a code smell of violating the Single responsibility principle (a class should do one thing). In reality I don't think the class actually violates this principle as the SSA is part of the  the Registration Request.

In fact, the SSA validation logic is in a different class and is a dependency of this class (which is great); so it's just the name that needs tweaking.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
```diff
@@ -0,0 +1,319 @@
...
+    RequestAndSsaSignatureValidationFilter(
+            RegistrationRequestObjectFromJwtSupplier registrationRequestObjectFromJwtSupplier,
+            Collection<String> supportedSigningAlgorithms,
+            DCRUtils dcrUtils,
+            DCRSsaSignatureValidator ssaValidator,
+            DCRRegistrationRequestJwtSignatureValidator registrationRequestJwtValidator) {
+        Reject.ifNull(registrationRequestObjectFromJwtSupplier, "RegistrationRequestObjectFromJwtSupplier " +
+                "must be provided");
+        Reject.ifNull(supportedSigningAlgorithms, "supportedSigningAlgorithms must be provided");
+        Reject.ifNull(registrationRequestJwtValidator, "registrationRequestJwtValidator must be provided");
```
dbadham-fr, line 99: validate that all of the dependencies are being provided.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRRegistrationRequestJwtSignatureValidatorJwks.java
```diff
@@ -0,0 +1,111 @@
...
+
+    private static final Logger log = LoggerFactory.getLogger(DCRRegistrationRequestJwtSignatureValidatorJwks.class);
+    private final JwtSignatureValidator jwtSignatureValidator;
+
+    /**
+     * Constructor
+     * @param jwtSignatureValidator service that is used to validate a SignedJwt against a JWKSet
+     */
+    public DCRRegistrationRequestJwtSignatureValidatorJwks(JwtSignatureValidator jwtSignatureValidator) {
+        this.jwtSignatureValidator = jwtSignatureValidator;
```
dbadham-fr, line 51: I'm sure this is just oversight as you are doing it in other places, but it is always a good practice to valid required dependencies are supplied in the constructor so we can fail fast here.

dbadham-fr (commented):
I'm slightly confused by the names of the following:
DCRRegistrationRequestJwtSignatureValidator
DCRRegistrationRequestJwtSignatureValidatorJwks
DCRRegistrationRequestJwtSignatureValidatorJwksUri

The latter 2 classes are named very similarly to the former but are in fact instances of: DCRRegistrationRequestValidator

I think it might be nicer for the DCRRegistrationRequestJwtSignatureValidatorJwks and DCRRegistrationRequestJwtSignatureValidatorJwksUri to be transformed into classes: EmbeddedJwksSupplier and JwksFromUriSupplier (names need some more thought...).

DCRRegistrationRequestJwtSignatureValidator is then responsible for doing the signature validation, delegating to the correct Supplier to obtain the Jwks to use. Then handling the success / error response for the validation is in one place only.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSignatureValidationException.java
```diff
@@ -0,0 +1,63 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.dcr.request;
+
+import org.forgerock.util.Reject;
+
+final public class DCRSignatureValidationException extends Exception{
```
dbadham-fr, line 20: Reuse com.forgerock.sapi.gateway.dcr.ValidationException

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/TrustedDirectoryOpenBankingTest.java
```diff
@@ -36,7 +36,7 @@ public class TrustedDirectoryOpenBankingTest implements TrustedDirectory {
      * The name of the claim in the Open Banking Test Directory issued software statement that holds the jwks_uri
      * against which certificates associated with this software statement may be validated
      */
-    final static String softwareJwksUriClaimName = "software_jwks_endpoint";
+    public final static String softwareJwksUriClaimName = "software_jwks_endpoint";
```
dbadham-fr, line 39: Why has this been made public? Callers should be using the getter

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
```diff
@@ -0,0 +1,319 @@
...
+            return ssaValidationPromise.thenAsync( response ->
+                registrationRequestJwtValidator.validateRegistrationRequestJwtSignature(
+                            fapiInteractionId, ssaClaimsSet, registrationRequestJwt)
+                            .thenAsync(regRequestValidationResponse -> {
+                        if (regRequestValidationResponse.getStatus() != Status.OK) {
+                            return Promises.newResultPromise(response);
+                        }
+                        return next.handle(context, request);
+                    }, ex -> {
+                        return Promises.newResultPromise(new Response(Status.BAD_REQUEST).setEntity(getJsonResponseBody(ex)));
```
dbadham-fr, line 138: When does this exception handler get called?

Also can you take a look at reformatting this async chain please? There are 2 checked exception handlers at the same indentation, making it hard to tell which async call they relate to.

It may be more readable to extract methods that have a sig which matches the org.forgerock.util.AsyncFunction, and then use a method reference, or have a method that returns a org.forgerock.util.AsyncFunction object. Then use this pattern when the chaining starts to get complex

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
```diff
@@ -0,0 +1,319 @@
...
+                    }), ex -> {
+                        log.debug("({}) Failed to get jwks from jwks_uri {}", fapiInteractionId, "caught exception: " + ex.getMessage(), ex);
+                        return Promises.newResultPromise(new Response(Status.BAD_REQUEST));
```
dbadham-fr, line 141: Review exception handling and logging here.

If you switch to using the com.forgerock.sapi.gateway.dcr.ValidationException then you can also make use of com.forgerock.sapi.gateway.dcr.ErrorResponseFactory to build the response objects which are OAuth 2.0 compliant, currently the response has an empty body.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSsaSignatureValidator.java
```diff
@@ -0,0 +1,155 @@
...
+import com.forgerock.sapi.gateway.dcr.utils.DCRUtils;
+import com.forgerock.sapi.gateway.jwks.JwkSetService;
+import com.forgerock.sapi.gateway.jws.JwtSignatureValidator;
+import com.forgerock.sapi.gateway.trusteddirectories.TrustedDirectory;
+import com.forgerock.sapi.gateway.trusteddirectories.TrustedDirectoryService;
+
+/**
+ * Validates the Signature of the software statement jwt.
+ */
+public class DCRSsaSignatureValidator {
```
dbadham-fr, line 44: Add some unit tests for this class please

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSsaSignatureValidator.java
```diff
@@ -0,0 +1,155 @@
...
+        JwtClaimsSet ssaClaimSet = ssa.getClaimsSet();
+
+        final String JWT_NAME = "software statement assertion";
+        String ssaIssuer = dcrUtils.getJwtIssuer(JWT_NAME, ssa);
+        log.debug("({}) {} issuer is {}", transactionId, JWT_NAME, ssaIssuer);
+
+        TrustedDirectory ssaIssuingDirectory = dcrUtils.getIssuingDirectory(trustedDirectoryService, ssaIssuer);
+
+        String issuingDirectoryJwksUriString = ssaIssuingDirectory.getDirectoryJwksUri();
+        return issuingDirectoryJwksUriFromString(transactionId, issuingDirectoryJwksUriString, ssaIssuer);
```
dbadham-fr, line 131: I think it makes sense to update the return type of com.forgerock.sapi.gateway.trusteddirectories.TrustedDirectory#getDirectoryJwksUri to URL, then we dodgy Trusted Directory config will be rejected during the configuration phase and we do not need to handle the case on the runtime hot path

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/utils/DCRUtils.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * get the value 'iss' claim, which hold a value indicating the issuer of the JWT.
+     * @param jwtName the name of the jwt from which the issuer is being obtained. Used in the error_description
+     *                of the exception thrown if the issuer can't be obtained
+     * @param signedJwt the signed JWT from which to obtain the issuer
+     * @return a string containing the issuer
+     * @throws DCRSignatureValidationException if the JWT does not contain an issuer claim
+     */
+    @NotNull
+    public String getJwtIssuer(String jwtName, SignedJwt signedJwt) throws DCRSignatureValidationException {
```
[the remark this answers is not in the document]
BohoCode: I prefer to pass a class into the constructor of dependent classes. I don't like statics as it make mocking/testing difficult.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/utils/JwtReconstructionException.java
```diff
@@ -0,0 +1,42 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.dcr.utils;
+
+/**
+ * Checked exception that can be thrown when reconstruction of a JWT from a b64 encoded JWT string fails
+ */
+public class JwtReconstructionException extends Exception {
```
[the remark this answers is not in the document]
BohoCode: The jose JwtReconstruction is a Runtime exception. 
I want this to be a checked exception so that I can know when the user has supplied a badly formatted jwt and write a good error about it. Unchecked makes sense if it's a config thing and affects system start up, e.g. the jwt is a part of our configuration, but when user supplied it's important to know when the third party library throws an exception. 

I think for our system a utility method that makes you have to handle the exception makes more sense.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSsaSignatureValidator.java
BohoCode: Ah, yes. I was thinking at one stage that IG did include a validator, but perhaps it didn't I sometimes find @NotNull good as a documentation mechanism, and of course we could always choose to put the hibernate validator on the class path and initialise the validator. 

However, I agree that I started off using @NotNull, but it's usage is inconsistent, so probably better off removed.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
BohoCode: Agreed. The naming was niggling me, but just wanted to get the PR created. Will rename.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSignatureValidationException.java
BohoCode: ValidationException is a runtime exception. I prefer to see the exceptions in the method signatures and be forced to handle them. Better for people following on. This gives a chance to make decent log entries that we know will have the transactionId in them.

When RuntimeExceptions happen and you search logs using tx id, then you don't see the runtime exceptions. You just see that processing stopped. THen you have to pin a log entry, remove the tx id from the search and search again. Then you see the runtime exception. 

This is not good for customers that want to diagnose why a TPPs method call failed.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RequestAndSsaSignatureValidationFilter.java
BohoCode: Ah, I meant to test all that and forgot. Will review and rationalise all that lump of handling.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/DCRSignatureValidationException.java
dbadham-fr: We can make ValidationException a checked exception, the class defs are otherwise identitcal.

I personally prefer RuntimeException which get caught at a suitable level with an Exception Handler i.e. in the Filter.filter method. As you need to think about handling unexpected RuntimeExceptions anyway, so I prefer to not have to pollute all of the sigs with throws ValidationException (and a list of other exceptions as the biz logic complexity increases).

But I'm happy to go down the checked exception path if you prefer.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/utils/DCRUtils.java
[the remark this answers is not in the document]
dbadham-fr: Passing dependencies via the constructor is definitely the way to go, however I don't see Utils classes as dependencies.

I'm not a big fan of these utils style classes but what you have here seems ok. It adheres to the convention of a utils class: stateless, immutable, side affect free, thread safe. Therefore there should be no need to ever mock this thing out.

For testing, you will need to provide valid test data / objects which these utils methods can operate on. I don't think this is a bad thing, I prefer to pass proper objects around in test rather than partially mocked stuff.

Otherwise, if you want to treat this as a true dependency, then the name needs some thought. Vaguely named classes tend to become dumping grounds for stuff, and as the DcrUtils object gets passed everywhere then it becomes a convenient way to fudge things.

BohoCode:
Example log entry on success:

image

BohoCode:
Example log entry on failure: 
image

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/utils/DCRUtils.java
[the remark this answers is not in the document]
dbadham-fr: Some more thoughts about this issue, I like that you are trying to follow a DI approach and supply the dependency, the issue I have is that there is a Utils class idiom in java and this class is not following it.

On the other hand, I don't particularly like the static Utils class idiom so it's probably not worthwhile making DcrUtils follow this.

Either way, having such a class seems to me to violate the Single Responsibility principle so should really be avoided.

Reviewing the class again I propose the following:

1. com.forgerock.sapi.gateway.dcr.utils.DCRUtils#getSignedJwt , this is not DCR specific and would be useful whenever we need to decode a base64 string representation of a JWT, therefore it should be moved into a JwtDecoder class in package: com.forgerock.sapi.gateway.jws
2. The remaining methods appear to relate to extracting data from the Registration Request, therefore I think we should create a RegistrationRequest object and a RegistrationRequestDecoder (which builds the object from the HTTP Request). Then we can reject upfront badly formed request and provide instance methods which extract the data we need

I think this is a much better OOP approach.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/SecureApiGatewayClassAliasResolver.java
```diff
@@ -45,6 +46,7 @@ public class SecureApiGatewayClassAliasResolver implements ClassAliasResolver {
         ALIASES.put("FetchApiClientJwksFilter", FetchApiClientJwksFilter.class);
         ALIASES.put("TransportCertValidationFilter", TransportCertValidationFilter.class);
         ALIASES.put("DefaultTransportCertValidator", DefaultTransportCertValidator.class);
+        ALIASES.put("RequestAndSsaSignatureValidationFilter", RegistrationRequestJwtSignatureValidationFilter.class);
```
dbadham-fr, line 49: Update alias

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/trusteddirectories/TrustedDirectorySecureApiGateway.java
```diff
@@ -38,7 +38,7 @@ public class TrustedDirectorySecureApiGateway implements TrustedDirectory {
 
     final static boolean softwareStatementHoldsJwksUri = false;
 
-    final static String softwareStatementJwksClaimName = "software_jwks";
+    public final static String softwareStatementJwksClaimName = "software_jwks";
```
dbadham-fr, line 41: should not be public

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/jwt/JwtException.java
```diff
@@ -0,0 +1,42 @@
...
+package com.forgerock.sapi.gateway.common.jwt;
+
+/**
+ * Exception thrown when issues occur processing jwts
+ */
+public class JwtException extends Exception {
+
+    public JwtException() {
+        super();
+    }
```
dbadham-fr, line 25: Remove this constructor, we always want a message

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/jwt/JwtException.java
```diff
@@ -0,0 +1,42 @@
...
+        super(message, cause);
+    }
+
+    public JwtException(Throwable cause) {
+        super(cause);
+    }
+
+    protected JwtException(String message, Throwable cause, boolean enableSuppression, boolean writableStackTrace) {
+        super(message, cause, enableSuppression, writableStackTrace);
+    }
```
dbadham-fr, line 41: Remove this, if we need an Exception that extends JwtException and needs this functionality then it can add it.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/rest/ContentTypeFormatterJson.java
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * Formats a Map of String key value pairs into a json string
+ */
+public class ContentTypeFormatterJson implements ContentTypeFormatter{
+    @Override
+    public String getFormattedResponse(Map<String, String> fields) {
+        if(fields == null){
+            return "{}";
+        }
+        ObjectMapper objectMapper = new ObjectMapper();
```
dbadham-fr, line 32: Reuse the objectMapper

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/rest/ContentTypeFormatterHtml.java
```diff
@@ -0,0 +1,45 @@
...
+ * Formats a map of String key value pairs into a representative HTML string
+ */
+public class ContentTypeFormatterHtml implements ContentTypeFormatter{
+    private static final Logger logger = LoggerFactory.getLogger(ContentTypeFormatterHtml.class);
+
+    public String getFormattedResponse(Map<String, String> errorForm) {
+        if(errorForm == null){
+            errorForm = Map.of();
+        }
+        StringBuilder errorMessageBuilder = new StringBuilder("<!doctype html><html lang=\"en\"><head><meta " +
```
dbadham-fr, line 33: call append rather than concat

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/rest/ContentTypeFormatterText.java
```diff
@@ -0,0 +1,41 @@
...
+public class ContentTypeFormatterText implements ContentTypeFormatter {
+
+    private static final Logger logger = LoggerFactory.getLogger(ContentTypeFormatterText.class);
+
+    @Override
+    public String getFormattedResponse(Map<String, String> fields) {
+        StringBuilder errorMessageBuilder = new StringBuilder();
+        fields.forEach((key, val) -> {
+            logger.debug("processing entry for {}, val is {}", key, val);
+            errorMessageBuilder.append(key + ": " + val + "\n");
```
dbadham-fr, line 35: call append rather than concat

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/rest/ContentTypeNegotiator.java
```diff
@@ -0,0 +1,103 @@
...
+    private final TreeMap<Float, List<String>> mediaTypesMap;
+    
+    private final List<String> supportedContentTypes;
+
+    public ContentTypeNegotiator(List<String> supportedContentTypes) {
+        Reject.ifNull(supportedContentTypes, "supportedContentTypes must not be null");
+        if(supportedContentTypes.isEmpty()){
+            throw new IllegalArgumentException("supportedContentTypes must not be empty");
+        }
+        mediaTypesMap = new TreeMap<>();
```
dbadham-fr, line 42: This doesn't look threadsafe.

The MultiValueMap are wrapping this map and modifying its contents per call to buildMapOfAcceptableMediaTypes.

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/models/SoftwareStatement.java
```diff
@@ -0,0 +1,95 @@
...
+
+import com.forgerock.sapi.gateway.common.jwt.ClaimsSetFacade;
+import com.forgerock.sapi.gateway.common.jwt.JwtException;
+
+public class SoftwareStatement {
+
+    private final SignedJwt signedJwt;
+    private final ClaimsSetFacade claimsSet;
+    private String org_id;
+    private String software_id;
```
dbadham-fr, line 32: use lowerCamel :camel:

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RegistrationRequestEntityValidatorFilter.java
```diff
@@ -0,0 +1,150 @@
...
+                    registrationRequest);
+            log.info("({}) created context attribute " + REGISTRATION_REQUEST_KEY, transactionId);
+            return next.handle(context, request);
+        } catch (DCRException exception){
+            List<String> acceptHeader = acceptHeaderSupplier.apply(context, request);
+            Response response = responseFactory.getResponse(transactionId, acceptHeader, Status.BAD_REQUEST,
+                    exception.getErrorFields());
+            log.info("({}) Failed to understand the Registration Request body: {}", transactionId,
+                    exception.getMessage(), exception);
+            return Promises.newResultPromise(response);
```
dbadham-fr, line 105: Sorry but I don't really see the use case for the content type negotiation for error responses.

Here are my concerns:

- [OAuth spec](https://www.rfc-editor.org/rfc/rfc7591#section-3.2.2) states that error messages are application/json
- Content negotiation is only done for the error path, if I were using an API I would expect it to apply to all responses
- I don't expect to see much usage of this feature, of the choices on offer json is the most popular.
  - Json is also easy to read by humans and machines (especially these simple error objects)
- It is adding extra complexity to an already large and complex PR

Appreciate that a lot of effort has gone in, especially to write tests but I really do question why we need this functionality.

I like that there is a common errorResponseFactory, this is an approach that we should continue to use to ensure consistent error responses are produced (and cut duplication).

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/common/jwt/SapiJwt.java
```diff
@@ -0,0 +1,120 @@
...
+    }
+
+    /**
+     * Set a flag that indicates if the data in the Software Statement can be trusted, i.e. if the signature has been
+     * validated against the relevant JWKS
+     * @param signatureHasBeenValidated true if the signature has been validated, false if it has not.
+     */
+    public void setSignatureHasBeenValidated(boolean signatureHasBeenValidated) {
+        this.signatureHasBeenValidated = signatureHasBeenValidated;
+    }
```
dbadham-fr, line 71: It would be better to manage this state inside this class, otherwise it is open to abuse as anyone can set the flag to true.

The nimbus lib does this, see classes: com.nimbusds.jwt.SignedJWT  / com.nimbusds.jose.JWSObject

Happy for this to be merged as is and fixed in another issue

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/sigvalidation/RegistrationRequestJwtSignatureValidationFilter.java
```diff
@@ -0,0 +1,226 @@
...
+        return "{\"error_code\":\"" + ex.getErrorCode().getCode() + "\"," +
+                "\"error_description\":\"" + ex.getErrorDescription() + "\"}";
+    }
+
+
+    /**
+     * Heaplet is used to get arguments from the IG config and create a RequestAndSsaSignatureValidationFilter
+     * on the IG heap.
+     */
+    public static class Heaplet extends GenericHeaplet {
```
dbadham-fr, line 160: Document how to create the filter via config, see com.forgerock.sapi.gateway.mtls.TransportCertValidationFilter.Heaplet for an example

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/dcr/request/RegistrationRequestEntityValidatorFilter.java
```diff
@@ -0,0 +1,162 @@
...
+                    Status.INTERNAL_SERVER_ERROR, errorFields);
+            return Promises.newResultPromise(response);
+        }
+    }
+
+    /**
+     * Heaplet is used to get arguments from the IG config and create a RequestAndSsaSignatureValidationFilter
+     * on the IG heap.
+     */
+    public static class Heaplet extends GenericHeaplet {
```
dbadham-fr, line 135: Document how to create the filter via config, see com.forgerock.sapi.gateway.mtls.TransportCertValidationFilter.Heaplet for an example

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/SecureApiGatewayClassAliasResolver.java
dbadham-fr: Alias doesn't match class name

dbadham-fr (approved):
Approved! Good job!

PR #275, opened 2023-01-11
637: vrp payment account

jorgesanchezperez:
- Updated pach request body to fix the account id and update the debtor account data when the decision is submitted
Issue: https://github.com/secureapigateway/secureapigateway/issues/637

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
@@ -36,6 +36,7 @@ def buildPatchRequest(incomingRequest, intentType) {
         ]);
     }
 
+    // Accounts Intent: Update accounts Ids
     if (incomingRequest.accountIds) {
         body.push([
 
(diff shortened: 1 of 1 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
```diff
@@ -46,46 +47,43 @@ def buildPatchRequest(incomingRequest, intentType) {
...
-            ])
-        }
-        if(incomingRequest.data.debtorAccount.identification) {
-            body.push([
-
-                    "operation": "add",
-                    "field"    : "OBIntentObject/Data/Initiation/DebtorAccount/Identification",
-                    "value"    : incomingRequest.data.debtorAccount.identification
-            ])
+        if(incomingRequest.data.debtorAccount.accounts[0]) {
```
dbadham-fr, line 52: is there a use case for debtor account containing an array?

jorgesanchezperez: Now the `debtorAccount` object is a `FRFinancialAccount` instead of a `FRAccountIdentifier` to update properly the accountId, and then we get the first account from the list that will contains only one account, the account selected by the PSU user when the decision submitted is `Authorised`.

BohoCode (approved):

PR #289, opened 2023-01-24
731: optional debtor account validation

jorgesanchezperez:
- Added new scriptable filter for payment initiation routes to validate the debtor account when is provided
- Added new groovy script to call RS to retrieve the account by identifiers provided in the debtor account consent 
> The consent will be rejected with an error description when the debtor account does not exist
```json
# response error example
{
    "Code": "400",
    "Id": "5412c3719b1535a0b93d9fef9322b95d",
    "Message": "[Status: 400 Bad Request]",
    "Errors": {
        "ErrorCode": "UK.OBIE.Field.Invalid",
        "Message": "Bad request [Invalid debtor account, the debtor account provided in the consent does not exist]"
    }
}
```
Issue: https://github.com/secureapigateway/secureapigateway/issues/731

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
@@ -1,0 +1,53 @@
+def fapiInteractionId = request.getHeaders().getFirst("x-fapi-interaction-id");
+if(fapiInteractionId == null) fapiInteractionId = "No x-fapi-interaction-id"
+SCRIPT_NAME = "[ValidateDebtorAccountInRS] (" + fapiInteractionId + ") - "
+
+logger.debug(SCRIPT_NAME + "Running...")
+def method = request.method
+
+
+switch(method.toUpperCase()) {
+
+    case "POST":
+        def debtorAccount = request.entity.getJson().Data.Initiation.DebtorAccount
+        if(debtorAccount !=null){
+            def requestURI = routeArgRsBaseURI + "/backoffice/accounts/search/findByAccountIdentifiers" +
+                    "?name=" + debtorAccount.Name + "&identification=" + debtorAccount.Identification +
+                    "&schemeName=" + debtorAccount.SchemeName
+
+            logger.debug(SCRIPT_NAME + " findByAccountIdentifiers raw request uri: " + requestURI)
+            Request RSRequest = new Request()
+            RSRequest.setUri(requestURI)
+            RSRequest.setMethod('GET')
+            if (request.headers.get("x-fapi-financial-id") != null) {
+                RSRequest.putHeaders(request.headers.get("x-fapi-financial-id"))
+            }
+
+            return http.send(RSRequest).thenAsync(RSResponse -> {
+                def RSResponseStatus = RSResponse.getStatus();
+                if (RSResponseStatus != Status.OK) {
+                    message = "Failed to find the debtor account by account identifiers"
+                    logger.error(SCRIPT_NAME + message)
+                    logger.error(SCRIPT_NAME + RSResponse.getEntity().getJson())
+                    def response = new Response(RSResponseStatus)
+                    response.status = RSResponseStatus
+                    response.entity = RSResponse.entity.getJson()
+                    return newResultPromise(response)
+                }
+                if (RSResponse.getEntity().isRawContentEmpty()) {
+                    message = "Invalid debtor account, the debtor account provided in the consent does not exist"
+                    logger.error(SCRIPT_NAME + message)
+                    response = new Response(Status.BAD_REQUEST)
+                    response.headers['Content-Type'] = "application/json"
+                    response.status = Status.BAD_REQUEST
+                    response.entity = "{ \"error\":\"" + message + "\"}"
+                    return newResultPromise(response)
+                }
+                return next.handle(context, request)
+            })
+        }
+    default:
+        logger.debug(SCRIPT_NAME + "Skipped")
+}
+
+next.handle(context, request)
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
@@ -16,25 +16,25 @@ switch(method.toUpperCase()) {
                     "&schemeName=" + debtorAccount.SchemeName
 
             logger.debug(SCRIPT_NAME + " findByAccountIdentifiers raw request uri: " + requestURI)
-            Request RSRequest = new Request()
-            RSRequest.setUri(requestURI)
-            RSRequest.setMethod('GET')
+            Request rsRequest = new Request()
+            rsRequest.setUri(requestURI)
+            rsRequest.setMethod('GET')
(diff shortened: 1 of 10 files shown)


dbadham-fr (commented):
Is there logic to validate that the DebtorAccount belongs to the PSU?

config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
```diff
@@ -0,0 +1,53 @@
...
+                    "?name=" + debtorAccount.Name + "&identification=" + debtorAccount.Identification +
+                    "&schemeName=" + debtorAccount.SchemeName
+
+            logger.debug(SCRIPT_NAME + " findByAccountIdentifiers raw request uri: " + requestURI)
+            Request RSRequest = new Request()
+            RSRequest.setUri(requestURI)
+            RSRequest.setMethod('GET')
+            if (request.headers.get("x-fapi-financial-id") != null) {
+                RSRequest.putHeaders(request.headers.get("x-fapi-financial-id"))
+            }
```
dbadham-fr, line 24: this should be able to be removed, you just need to configure the ScriptableFilter to use the `ForgeRockClientHandler`

dbadham-fr: Ah, I thought this was the `x-fapi-interaction-id`

Why do we want to send the `x-fapi-financial-id` this has been removed from the OB spec

jorgesanchezperez:
> Is there logic to validate that the DebtorAccount belongs to the PSU?

In this request point the PSU is unknown.

The endpoint accepts as optional the userId (PSU) and this value will be used also to find the account  by userId and identifiers if the userId is not null otherwise the find account method will be only by account identifiers.
```java
ResponseEntity<FRAccountWithBalance> getAccountWithBalanceByIdentifiers(
            @ApiParam(value = "The ID of the PSU who owns the account. (non mandatory)")
            @RequestParam(value = "userId", required = false) String userId,
            @ApiParam(value = "Indicates the name of account identifier to filter the account identifiers. Defaults to false.")
            @RequestParam(value = "name", defaultValue = "false") String accountIdentifierName,
            @ApiParam(value = "Indicates the identification of account identifier to filter the account identifiers. Defaults to false.")
            @RequestParam(value = "identification", defaultValue = "false") String accountIdentifierIdentification,
            @ApiParam(value = "Indicates the schema name of account identifier to filter the account identifiers. Defaults to false.")
            @RequestParam(value = "schemeName", defaultValue = "false") String accountIdentifierSchemeName
    );
```

config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
jorgesanchezperez: It will be better create an issue to update the config to configure the `ScriptableFilter` to use the `ForgerRockClientHandler` and make a clean up through all existing scripts.
What is your thoughts?

dbadham-fr:
> In this request point the PSU is unknown.

Yeah that makes sense.

So where will the validation that the PSU owns the account go? In the RCS?

As if that's the case then we may as well defer doing the debtor account validation until then, they can create the Intent but then it will get Rejected when they try to authorise it (if they don't own the account or the account does not exist)

Then it saves making an additional REST API call

jorgesanchezperez:
> So where will the validation that the PSU owns the account go? In the RCS?

Yes, there is another validation when the consent details is building to be displayed the consent in RCS-UI.

> As if that's the case then we may as well defer doing the debtor account validation until then, they can create the Intent but then it will get Rejected when they try to authorise it (if they don't own the account or the account does not exist)

I don't like to accept a consent with a invalid DebtorAccount field because:
- Does not make sense stored consents where the debtor account is wrong.
- If we accept the consent with a wrong debtor account we are increasing the exceptions when the customer/asps/tpp request later flows.
- Make sense restrict the errors from the origin instead of reject the consent or throw and error when the consent will be displayed in RCS-UI

> Then it saves making an additional REST API call

Yes, but:
The resource cost to call RS when the debtor account is provided in the payment initiation is low and the number of consents that raise an error later in other flows could be high if we do not restrict these objects validating them in the first step flow.

dbadham-fr:
Yeah I don't like the idea of storing junk accounts either, but you will still have the situation where a PSU can create a valid consent with a debtor account that exists and belongs to someone else.

As long as we are doing the PSU ownership validation in the RCS then I'm happy for this initial sanity validation to be added.

I think it's one of those things that becomes much simpler when we move the consent processing out of the gateway, then validating that an account is one that exists within the ASPSP is much more trivial.

config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
dbadham-fr: May as well fix this new filter that your adding then tackle any other calls to the backoffice API later if you think those are broken, as it is a quick fix.

config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+    case "POST":
+        def debtorAccount = request.entity.getJson().Data.Initiation.DebtorAccount
+        if(debtorAccount !=null){
+            def requestURI = routeArgRsBaseURI + "/backoffice/accounts/search/findByAccountIdentifiers" +
+                    "?name=" + debtorAccount.Name + "&identification=" + debtorAccount.Identification +
+                    "&schemeName=" + debtorAccount.SchemeName
+
+            logger.debug(SCRIPT_NAME + " findByAccountIdentifiers raw request uri: " + requestURI)
+            Request RSRequest = new Request()
```
dbadham-fr, line 19: fix variable name to lowerCamel, same for RSResponseStatus

jorgesanchezperez: done

config/7.1.0/securebanking/ig/scripts/groovy/ValidateDebtorAccountInRS.groovy
```diff
@@ -0,0 +1,53 @@
...
+            logger.debug(SCRIPT_NAME + " findByAccountIdentifiers raw request uri: " + requestURI)
+            Request rsRequest = new Request()
+            rsRequest.setUri(requestURI)
+            rsRequest.setMethod('GET')
+            if (request.headers.get("x-fapi-financial-id") != null) {
+                rsRequest.putHeaders(request.headers.get("x-fapi-financial-id"))
+            }
+
+            return http.send(rsRequest).thenAsync(rsResponse -> {
+                def RSResponseStatus = rsResponse.getStatus();
```
dbadham-fr, line 27: :camel:

jorgesanchezperez: Ups! done!

dbadham-fr (dismissed):

dbadham-fr (approved):

PR #293, opened 2023-01-26
736: Allow only Sweeping VRPType

frlaura-jianu:
Issue: https://github.com/secureapigateway/secureapigateway/issues/736
Description: Added a new filter to the consent route to check if the VRPType is "Sweeping".

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/VRPTypeVerifier.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/VRPTypeVerifier.groovy
@@ -1,0 +1,36 @@
+import org.forgerock.http.protocol.*
+import com.forgerock.sapi.gateway.rest.HttpHeaderNames
+
+String fapiInteractionId = request.getHeaders().getFirst(HttpHeaderNames.X_FAPI_INTERACTION_ID)
+if (fapiInteractionId == null) {
+    fapiInteractionId = 'No ' + HttpHeaderNames.X_FAPI_INTERACTION_ID + ' header'
+}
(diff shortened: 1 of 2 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/VRPTypeVerifier.groovy
```diff
@@ -0,0 +1,36 @@
...
+
+def vrpAllowed = false
+
+for (String vrpType : vrpTypeFromRequest) {
+    logger.debug(SCRIPT_NAME + 'vrpType: ' + vrpType)
+    if (vrpType == 'UK.OBIE.VRPType.Sweeping') {
+        vrpAllowed = true;
+        break;
+    }
+}
```
jorgesanchezperez, line 23: If the platform support only `UK.OBIE.VRPType.Sweeping` we need to reject the consent if any other VRPtype is provided in that list, for example:
VRPType = ["`UK.OBIE.VRPType.Sweeping`", "`UK.OBIE.VRPType.Other`"] reject the consent because we support only Sweeping.

In short the list need to contains only `UK.OBIE.VRPType.Sweeping`
VRPType = ["`UK.OBIE.VRPType.Sweeping`"]

jorgesanchezperez: Probably the for could be deleted and use a simple conditional.

jorgesanchezperez (changes requested):
Review the VRPType list check.

jorgesanchezperez (approved):

PR #299, opened 2023-02-06
715: X-fapi-financial-id header should be optional

frlaura-jianu:
Issue: https://github.com/secureapigateway/secureapigateway/issues/715
Description: Changed x-fapi-financial-id header requirement from mandatory to optional.

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/CalculateResponseElementsInRS.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/CalculateResponseElementsInRS.groovy
@@ -22,8 +22,8 @@ switch(method.toUpperCase()) {
         RSRequest.setUri(requestURI)
         RSRequest.setMethod('POST')
         RSRequest.setEntity(request.entity.getJson())
-        if (request.headers.get("x-fapi-financial-id") != null) {
-            RSRequest.putHeaders(request.headers.get("x-fapi-financial-id"))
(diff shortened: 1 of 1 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/CalculateResponseElementsInRS.groovy
```diff
@@ -22,8 +23,8 @@ switch(method.toUpperCase()) {
         RSRequest.setUri(requestURI)
         RSRequest.setMethod('POST')
         RSRequest.setEntity(request.entity.getJson())
-        if (request.headers.get("x-fapi-financial-id") != null) {
-            RSRequest.putHeaders(request.headers.get("x-fapi-financial-id"))
+        if (request.headers.get(X_FAPI_FINANCIAL_ID_HEADER) == null) {
```
dbadham-fr, line 26: This block just needs to be removed.

x-fapi-financial-id is no longer part of the OB spec.

Also update the issue to remove the points about it being optional.

The x-fapi-financial-id is the identifier of a bank, so can't be generated.
There was confusion in the past about this header and `x-fapi-interaction-id` which is a uuid.

frlaura-jianu: Removed the x-fapi-financial-id header check according to the spec.

dbadham-fr (approved):

PR #301, opened 2023-02-07
793: Protecting access to RCS-UI

dbadham-fr:
Adding routes which protect access to the RCS-UI. 

Route: 81-rcs-ui.json protects the call to the rcs-ui /consent endpoint and does authentication + authorisation checks.

Route: 82-rcs-ui-assets.json protects calls to the rcs-ui scripts and assets, it only authentication checks. Capturing is also turned off on this route to avoid spamming the logs with javascript/css files.

Utilising IdTokenValidationFilter to validate the consent_request JWT, this filter verifies the signature and validates the "exp", "iat", "iss" and "aud" claims.

The ConsentRequestAccessAuthorisationFilter uses the SsoTokenContext and JwtValidationContext set by the SingleSignOnFilter and IdTokenValidationFilter respectively. Using these contexts, the filter verifies that the userId of the SSOToken matches that of the consent_request JWT signed by AM, if these match then the filter allows the request to continue along the chain, otherwise a HTTP 401 is returned.

https://github.com/SecureApiGateway/SecureApiGateway/issues/793

### DIFF
--- a/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/consent/ConsentRequestAccessAuthorisationFilter.java
+++ b/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/consent/ConsentRequestAccessAuthorisationFilter.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright © 2020-2022 ForgeRock AS (obst@forgerock.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.consent;
+
+import java.util.Map;
+
+import org.forgerock.http.Filter;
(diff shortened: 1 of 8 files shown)


secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/consent/ConsentRequestAccessAuthorisationFilter.java
```diff
@@ -0,0 +1,181 @@
...
+     * occurs then it indicates a misconfiguration in a ForgeRock application.
+     */
+    static ExceptionHandler createDefaultExceptionHandler() {
+        return (ctxt, ex) -> {
+            LOGGER.warn("(" + FAPIUtils.getFapiInteractionIdForDisplay(ctxt) + ") Failed to get userId data required to do authorisation check", ex);
+            return new Response(Status.INTERNAL_SERVER_ERROR);
+        };
+    }
+
+    private final ExceptionHandler exceptionHandler;
```
dbadham-fr, line 89: This is a hook for testing purposes, it allows the unit tests to supply an ExceptionHandler instance which captures the exceptions passed to it. We can then do validation on these exceptions to ensure that a HTTP 500 response was triggered by a particular exception.

The 500 response should only occur if AM or IG change their impl of either the consent_request JWT or SsoToken format. There is intentionally no information returned to the end user as there is no action they can reasonably take, we want to avoid leaking any information at this point as another reason for a failure is someone attempting to hack this flow.

dbadham-fr:
This PR is blocked from being merged until: https://github.com/SecureApiGateway/SecureApiGateway/issues/801 is fixed.

The work has been tested using a hacked RCS-UI docker image

BohoCode:
Looks good

jorgesanchezperez (approved):

PR #306, opened 2023-02-10
794: Using internal RS k8s service address for IG reverse proxy calls to RS

dbadham-fr:
https://github.com/SecureApiGateway/SecureApiGateway/issues/794


dbadham-fr:
Closing, will create a new PR with a clean diff.

PR #309, opened 2023-02-14
592: idempotency validation

jorgesanchezperez:
- Added validation of idempotency requests for payment consents and file submission
Issue: https://github.com/secureapigateway/secureapigateway/issues/592

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
@@ -48,7 +48,7 @@ switch (method.toUpperCase()) {
         }
         // current time to filter the idempotency key expiration time (idempotencyKeyExpiration stored in seconds = creation time + 24hours)
         Instant currentInstantTime = Instant.now()
-        logger.debug(SCRIPT_NAME + "apiClientId: " + apiClientId + ", idempotencyKeyExpire: " + idempotencyKeyExpire +", in seconds: " + idempotencyKeyExpire.getEpochSecond())
+        logger.debug(SCRIPT_NAME + "apiClientId: " + apiClientId + ", current Instant: " + currentInstantTime +", in seconds: " + currentInstantTime.getEpochSecond())
         // filter by idempotency key, idempotency key expiration time (only valid for 24hours since has been created) and Oauth2ClientId
         def filter = "_queryFilter="+ idempotencyArgFieldToFilter + "+eq+%22" +
                 URLEncoder.encode(idempotencyKeyHeaderValue, "UTF-8") + "%22" +
@@ -78,6 +78,7 @@ switch (method.toUpperCase()) {
 
             def intentResponseResult = intentResponse.entity.getJson().result
             if (!intentResponseResult.isEmpty()) {
+                logger.info("Found a intent for Oauth2ClientId: " + apiClientId + " with not expired " + idempotencyArgFieldToFilter +": " + idempotencyKeyHeaderValue)
                 // For file submission the response needs to be Ok with empty entity
                 if(idempotencyArgFieldToFilter == "IdempotencyKeyFile") {
                     intentResponse.status = Status.OK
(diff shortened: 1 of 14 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
```diff
@@ -0,0 +1,70 @@
...
+// response object
+responseCheck = new Response(Status.OK)
+responseCheck.headers['Content-Type'] = "application/json"
+
+String idempotencyKeyHeaderValue = request.getHeaders().getFirst("x-idempotency-key")
+logger.debug(SCRIPT_NAME + "IdempotencyKey header value: " + idempotencyKeyHeaderValue)
+
+def method = request.method
+
+def filter = "_queryFilter="+ idempotencyArgFieldToFilter + "+eq+%22" + URLEncoder.encode(idempotencyKeyHeaderValue, "UTF-8") + "%22"
```
dbadham-fr, line 22: The key only needs to be unique to the TPP, not globally unique. Otherwise values sent by different TPPs could clash, we can't assume how they will implement the idempotency key generation.

dbadham-fr (changes requested):

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
jorgesanchezperez: If you mean the argument `idempotencyArgFieldToFilter` is not the key the key sent by the TPP is `URLEncoder.encode(idempotencyKeyHeaderValue, "UTF-8")` obtained from the request few lines above, `idempotencyArgFieldToFilter` is the field name where we store the key send by the TPP, for example for any payment initiation we store the key in the field `IdempotencyKey`, the file payment initiation is a little bit different because the submission of the file content need to be idempotent too, for that reason in that case we have two different fields, one for the consent named `IdempotencyKey` (the same as other initiations) and the specific field named `IdempotencyKeyFile` to store the idempotency key send by the TPP when the file content is submitted.

dbadham-fr: I may be misunderstanding the logic somewhere but here is an example scenario which I think will behave incorrectly with the current impl.

- dan-tpp apiClient sends a DomesticPaymentConsent with idempotencyKey = "1"
- jordi-tpp apiClient sends a DomesticPaymentConsent with idempotencyKey = "1"

What response does jordi-tpp get? From my reading of the code then you will get dan-tpp's consent sent back to you.

Each TPP's idempotencyKey must be treated as an independent set of values.

jorgesanchezperez: Got It! we need add to the filter the apiClient too to request only the consents linked with the TPP to avoid retrieve the consent of another TPP filtering only by idempotency-key.

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
```diff
@@ -0,0 +1,95 @@
...
+switch (method.toUpperCase()) {
+    case "POST":
+        if (idempotencyKeyHeaderValue == null || idempotencyKeyHeaderValue == "") {
+            message = "Failed to get create the resource, 'x-idempotency-key' header / value expected"
+            logger.error(SCRIPT_NAME + message)
+            responseCheck.status = Status.BAD_REQUEST
+            responseCheck.entity = "{ \"error\":\"" + message + "\"}"
+            return responseCheck
+            // the idempotencyKey size must be at most 40 characters
+        } else if(idempotencyKeyHeaderValue.length() > 39){
```
dbadham-fr, line 33: `> 40`

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
```diff
@@ -0,0 +1,95 @@
...
+        return http.send(intentRequest).thenAsync(intentResponse -> {
+            intentRequest.close()
+            logger.debug(SCRIPT_NAME + "Back from IDM")
+            def intentResponseStatus = intentResponse.getStatus()
+            if (intentResponseStatus != Status.OK) {
+                message = "Failed to validate idempotent request"
+                logger.error(SCRIPT_NAME + message)
+                response.status = intentResponseStatus
+                response.entity = "{ \"error\":\"" + message + "\"}"
+                return response
```
dbadham-fr, line 76: need to wrap with newResultPromise

dbadham-fr (changes requested):

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
jorgesanchezperez: Done!

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
jorgesanchezperez: Fixed condition and all returns with newResultPromise

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
jorgesanchezperez: done

config/7.1.0/securebanking/ig/scripts/groovy/CheckIdempotencyKey.groovy
```diff
@@ -0,0 +1,95 @@
...
+                response.status = intentResponseStatus
+                response.entity = "{ \"error\":\"" + message + "\"}"
+                return newResultPromise(response)
+            }
+
+            def intentResponseResult = intentResponse.entity.getJson().result
+            if (!intentResponseResult.isEmpty()) {
+                logger.info("Found a intent for Oauth2ClientId: " + apiClientId + " with not expired " + idempotencyArgFieldToFilter +": " + idempotencyKeyHeaderValue)
+                // For file submission the response needs to be Ok with empty entity
+                if(idempotencyArgFieldToFilter == "IdempotencyKeyFile") {
```
dbadham-fr, line 83: update

jorgesanchezperez: Fixed.

dbadham-fr (approved):

PR #316, opened 2023-02-27
831: Validate MTLS client cert when accessing token_endpoint

dbadham-fr:
Created TokenEndpointTransportCertValidationFilter which does mTLS certificate validation for the AM token endpoint route. This validation is deferred until the response path, so that we know that we have an authenticated client.

The new filter implements the same validation as TransportCertValidationFilter but is specialised to the response path. This means that it needs to lookup the resources to do the validation itself, rather than access them via the Context.

Logic in existing filters that fetch resources have been refactored to move them to their own service classes. These services are then plugged into the existing filters and reused in the new TokenEndpointTransportCertValidationFilter.

In this PR we are deprecating: CertificateFromHeaderSupplier , an issue has been raised to remove it: https://github.com/SecureApiGateway/SecureApiGateway/issues/856

https://github.com/SecureApiGateway/SecureApiGateway/issues/831

### DIFF
--- a/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/mtls/CertificateResolver.java
+++ b/secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/mtls/CertificateResolver.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright © 2020-2022 ForgeRock AS (obst@forgerock.com)
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.forgerock.sapi.gateway.mtls;
+
+import java.security.cert.CertificateException;
+import java.security.cert.X509Certificate;
+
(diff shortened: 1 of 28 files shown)


secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/mtls/TokenEndpointTransportCertValidationFilter.java
```diff
@@ -0,0 +1,278 @@
...
+/**
+ * Filter to validate that the client's MTLS transport certificate is valid when making a request to the Authorisation
+ * Server's token endpoint.
+ *
+ * This is a specialised version of {@link TransportCertValidationFilter}, it does the same
+ * validation, but has been adapted to do its validation on the response path. By deferring the validation to the response
+ * path then we can be sure that we have an authenticated client.
+ *
+ * The access_token returned by the Authorisation Server is inspected to retrieve the client_id by using the
+ * configurable accessTokenClientIdClaim. Once the client_id has been retrieve, it is used to look up the resources
```
BohoCode, line 63: Sp: Once the client_id has been retrieve**d**

secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/mtls/CertificateResolver.java
```diff
@@ -0,0 +1,40 @@
...
+import java.security.cert.X509Certificate;
+
+import org.forgerock.http.protocol.Request;
+import org.forgerock.services.context.Context;
+
+/**
+ * Resolves of a client's mTLS certificate
+ */
+public interface CertificateResolver {
+
```
BohoCode, line 28: Resolve indicates a decision is being made. CertificateRetriever or CertificateFetcher might be better?

BohoCode (approved):

BohoCode (approved):
Proper Job!

PR #322, opened 2023-03-01
691: Fixing ProcessDetachedSig header claim: http://openbanking.org.uk/iss validation

dbadham-fr:
http://openbanking.org.uk/iss header claim is validated to check it matches "{{org-id}}/{{software-statement-id}}"

To do this we need to get the ApiClient data, as the ProcessDetachedSig filter runs after FetchApiClientFilter then this data is available in the attributes context.

Fixing bug in IDM query which was not retrieiving the fields within the apiClientOrg object.

Fixing bug in ProcessDetachedSig which caused validation errors to be ignored as the isJwsSignatureValid was returning a Promise wrapping a boolean instead of a boolean.

https://github.com/SecureApiGateway/SecureApiGateway/issues/691

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ProcessDetachedSig.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ProcessDetachedSig.groovy
@@ -29,6 +29,8 @@ JWS spec: https://www.rfc-editor.org/rfc/rfc7515#page-7
 
  If ASPSPs have updated to v3.1.4 or later, they must not include the b64 claim in the header,
  and any TPPs using these ASPSPs must do the same.
+
+ This script relies on the apiClient attribute being set, therefore must be installed after the FetchApiClientFilter
  */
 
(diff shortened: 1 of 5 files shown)


config/7.1.0/securebanking/ig/scripts/groovy/ProcessDetachedSig.groovy
```diff
@@ -232,7 +234,7 @@ def isJwsSignatureValid(String detachedSignatureValue, RSAPublicKey rsaPublicKey
     boolean criticalParamsValid = validateCriticalParameters(jwsHeader);
     if (!criticalParamsValid) {
         logger.error(SCRIPT_NAME + "Critical params validations failed. Stopping further validations.")
-        return newResultPromise(false)
+        return false
```
dbadham-fr, line 237: very subtle bug, something to be careful with.

Due to groovy true: https://groovy-lang.org/semantics.html#the-groovy-truth , Promise<Boolean>(false) evaluates to true

BohoCode (approved):

PR #328, opened 2023-03-02
614: Prevent deleted ApiClients being retrieved by IdmApiClientService

dbadham-fr:
If the ApiClient has been deleted then raise an exception in IdmApiClientService.

 There is currently no use case for fetching deleted clients. Ensuring that we do not retrieve them will prevent errors occurring when a deleted client attempts to do an action.

https://github.com/SecureApiGateway/SecureApiGateway/issues/614


BohoCode:
Ace!

BohoCode (approved):

PR #342, opened 2023-03-08
859: Check route role against roles in ApiClient

BohoCode:
The CertificateCheck.groovy script used the roles from the presented eidas certificate. This restricts the API client to those that have an eidas style certificate. We now use the roles specified in the software statement at teh time the client performed dynamic client registration to determine the roles that the client is allowed to perform.

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/859


dbadham-fr (approved):

dbadham-fr (changes requested):
I've tested the PR in my env and func tests are failing

dbadham-fr (approved):

PR #343, opened 2023-03-13
885: Signing Kid env var updated to OB_ASPSP_SIGNING_KID

dbadham-fr:
https://github.com/SecureApiGateway/SecureApiGateway/issues/885


dbadham-fr:
Ooops, wrong field.

PR #349, opened 2023-03-16
812: release action

jorgesanchezperez:
Provides a way to build the release for the artifact ig-extensions calling a reusable workflow that lives in other repository in the same organisation
- Update release workflow 

As we decide to delete the `scm` and `distributionManagement` pom sections:
- Deleted from pom `scm` and `distributionManagement` sections
- Updated merge master workflow
- Update pr workflow

- Issue: https://github.com/secureapigateway/secureapigateway/issues/812

### DIFF
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -38,6 +38,9 @@ jobs:
       - name: Build maven
         run: |
           mvn clean deploy -DaltDeploymentRepository=release::default::"https://${{ secrets.FR_ARTIFACTORY_USER }}:${{ secrets.FR_ARTIFACTORY_TOKEN }}@maven.forgerock.org/artifactory/community/"
+        env:
+          MAVEN_USERNAME: ${{ secrets.FR_ARTIFACTORY_USER }}
+          MAVEN_CENTRAL_TOKEN: ${{ secrets.FR_ARTIFACTORY_TOKEN }}
 
       - name: docker build
         run: |
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -34,6 +34,9 @@ jobs:
           java-version: '17'
           architecture: x64
           cache: 'maven'
+          server-id: forgerock-private-releases # Value of the distributionManagement/repository/id field of the pom.xml
+          server-username: MAVEN_USERNAME # env variable for username in deploy
+          server-password: MAVEN_CENTRAL_TOKEN # env variable for token in deploy
 
       - name: Build maven
         run: |
--- a/.github/workflows/merge.yml
+++ b/.github/workflows/merge.yml
@@ -26,33 +26,25 @@ jobs:
       - run: |
           gcloud auth configure-docker
 
-      - name: Setup defaults action java and maven properties
-        uses: actions/setup-java@v1
+      - uses: actions/setup-java@v3
+        id: maven
+        name: set java and maven cache
         with:
-          java-version: "17"
+          distribution: 'adopt'
+          java-version: '17'
           architecture: x64
-          server-id: forgerock-private-releases # Value of the distributionManagement/repository/id field of the pom.xml
-          server-username: MAVEN_USERNAME # env variable for username in deploy
-          server-password: MAVEN_CENTRAL_TOKEN # env variable for token in deploy
-
-      - name: Cache Maven packages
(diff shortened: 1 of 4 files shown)


.github/workflows/merge.yml
```diff
@@ -26,33 +26,31 @@ jobs:
...
-      - name: Cache Maven packages
-        uses: actions/cache@v1
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven2-${{ hashFiles('**/pom.xml') }}
-
       - name: Build maven
         run: |
-          mvn clean install
+          mvn clean deploy -DaltDeploymentRepository=release::default::"https://${{ secrets.FR_ARTIFACTORY_USER }}:${{ secrets.FR_ARTIFACTORY_TOKEN }}@maven.forgerock.org/artifactory/community/"
```
BohoCode, line 43: Why are we deploying this to artifactory? Is this necessary?

.github/workflows/pr.yml
```diff
@@ -29,21 +29,18 @@ jobs:
...
           server-id: forgerock-private-releases # Value of the distributionManagement/repository/id field of the pom.xml
           server-username: MAVEN_USERNAME # env variable for username in deploy
           server-password: MAVEN_CENTRAL_TOKEN # env variable for token in deploy
 
-      - name: Cache Maven packages
-        uses: actions/cache@v1
-        with:
-          path: ~/.m2/repository
-          key: ${{ runner.os }}-maven2-${{ hashFiles('**/pom.xml') }}
-
```
BohoCode, line 46: Why remove this? Is it necessary for the release action to work?

.github/workflows/release.yml
```diff
@@ -1,16 +1,165 @@
...
+        id: checkout
+        name: checkout master
+        with:
+          ref: ${{ needs.release_prepare.outputs.release_branch }}
+
+      # https://github.com/crazy-max/ghaction-import-gpg
+      # https://httgp.com/signing-commits-in-github-actions/
+      - name: Import GPG key
+        id: gpg
+        uses: crazy-max/ghaction-import-gpg@v5
```
BohoCode, line 71: crazy-max?? Have we reviewed this code? Is it secure?

.github/workflows/release.yml
```diff
@@ -1,16 +1,165 @@
...
         with:
-          ref: ${{ github.ref }}
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY_BOT }}
+          passphrase: ${{ secrets.GPG_KEY_PASSPHRASE_BOT }}
```
BohoCode, line 74: Who's GPG account is this?

Why can't we just do a gpg --keyserver  --recv-keys <kid> instead of using this dodgy sounding gh action?

.github/workflows/release.yml
```diff
@@ -1,16 +1,165 @@
...
+          ref: ${{ needs.release_prepare.outputs.release_branch }}
+
+      # https://github.com/crazy-max/ghaction-import-gpg
+      # https://httgp.com/signing-commits-in-github-actions/
+      - name: Import GPG key
+        id: gpg
+        uses: crazy-max/ghaction-import-gpg@v5
         with:
-          ref: ${{ github.ref }}
+          gpg_private_key: ${{ secrets.GPG_PRIVATE_KEY_BOT }}
```
BohoCode, line 73: Not sure I would trust crazy-max with my private keys??

BohoCode (commented):
Generally I like the approach. I worry about a few of the security implications and third party actions being used though....

.github/workflows/merge.yml
jorgesanchezperez: This is the process to deploy the snapshot version of ig-extensions, until now we never publish it to forgerock maven repository, to follow the same cycle of any other projects I decided to publish that snapshot library too.

.github/workflows/pr.yml
jorgesanchezperez: No the maven cache is provided by `actions/setup-java@v3` 
```yaml
      - uses: actions/setup-java@v3
        id: maven
        name: set java and maven cache
        with:
          distribution: 'adopt'
          java-version: '17'
          architecture: x64
          cache: 'maven'
```
We no need anymore a specific step to create the maven cache.

PR #358, opened 2023-03-28
916: Fixing TokenEndpointTransportCertValidationFilter error responses

dbadham-fr:
When mTLS cert validation is done as part of the token_endpoint AS route, then error responses need to conform to the OAuth 2.0 spec. https://www.rfc-editor.org/rfc/rfc6749#section-5.2

Adding an errorResponse method to TokenEndpointTransportCertValidationFilter which produces errors in the correct format. Example response:
```
HTTP 401 UNAUTHORIZED
{
  "error":"invalid_client",
  "error_description":"Client mTLS certificate not provided"
}
```

The TransportCertValidationFilter is used on Open Banking routes, it needs to produce an error that conforms to the OB spec and and therefore has not been updated.

https://github.com/SecureApiGateway/SecureApiGateway/issues/916


dbadham-fr:
I've tested this change using the FAPI conformance suite running against my environment, test: `fapi1-advanced-final-ensure-mtls-holder-of-key-required` now passes

jorgesanchezperez (approved):

PR #360, opened 2023-03-29
883: Configuring ApiClient.name to be equal to a value from the software statement

dbadham-fr:
Adding support for clientName claim in the TrustedDirectory configuration, using this to extract the clientName from the software statement during DCR. Using the value extracted to set the IDM ApiClient.name field.

https://github.com/SecureApiGateway/SecureApiGateway/issues/883


dbadham-fr:
image

Tested this and we now see the TPP's name rather than software_id in the UI.

jorgesanchezperez (approved):

PR #381, opened 2023-04-27
947: Validation of initiation object, skip patch the debtor account and update accountId when debtor account is null

jorgesanchezperez:
Currently to store the account to perform a payment we use the field defined as `debtorAccount`.

Spec validation rule: The PISP must ensure that the Initiation and Risk sections of the international-payment match the corresponding Initiation and Risk sections of the international-payment-consent resource. If the two do not match, the ASPSP must not process the request and must respond with a 400 (Bad Request)

The `debtorAccount` field is patched when the psu submit the consent decision, in consequence when the TPP submit a payment, and the system validates the initiation object from payment request against the initiation object from the consent, the validation fails because the initiation objects do not match because the consent has been modified by the system.

We need to skip update the `debtorAccount` field in the consent and store the account to perform the payment in the `accountId` field outside the `OBIntentObject`.

- The debtor account must still the same provided by the TPP when creates the consent
- Deleted update the debtor account on the consent
- Added logic to set the accountId with the account to perform the payment

Issue: https://github.com/secureapigateway/secureapigateway/issues/947
Comment: https://github.com/secureapigateway/secureapigateway/issues/947#issuecomment-1521744538

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/RepoConsent.groovy
@@ -47,52 +47,14 @@ def buildPatchRequest(incomingRequest, intentType) {
         ]);
     }
 
-    // Payments Intent: Update Debtor account and accountId
-    if (incomingRequest.data && incomingRequest.data.debtorAccount) {
-        if(incomingRequest.data.debtorAccount.accounts[0]) {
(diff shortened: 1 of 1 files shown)


BohoCode (changes requested):
Please explain in the description why these changes were required, what problem these lines caused and your motivation behind changing them.

jorgesanchezperez:
> Please explain in the description why these changes were required, what problem these lines caused and your motivation behind changing them.

done

PR #385, opened 2023-05-15
Adding crosscheck between request consentId vs token consentId

mariantiris:
Issue: https://github.com/secureapigateway/secureapigateway/issues/979

Added a new validation script on the payment submit routes.
This script is comparing the consentId from the request payload versus the consentId from the provided access token. By doing this comparison, we prevent resources being submitted with a token obtained for another consent.

### DIFF
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
@@ -1,6 +1,5 @@
 import org.forgerock.http.protocol.*
 import com.forgerock.sapi.gateway.rest.HttpHeaderNames
-import groovy.json.JsonSlurper
 
 /**
  * This script is comparing the consentId from the request payload versus the consentId from the provided access token.
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
@@ -16,7 +16,7 @@ logger.debug(SCRIPT_NAME + 'Running...')
 def accessTokenIntentId = attributes.openbanking_intent_id
 
 if (!accessTokenIntentId) {
-    throw new IllegalStateException("openbanking_intent_id claim is missing from the access token");
+    throw new IllegalStateException("openbanking_intent_id claim is missing from the attributes context");
 }
 
 // Get the intent Id from the request body
--- a/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
+++ b/config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
@@ -15,19 +15,23 @@ logger.debug(SCRIPT_NAME + 'Running...')
 // Get the intent id from the access token
 def accessTokenIntentId = attributes.openbanking_intent_id
 
+if (!accessTokenIntentId) {
+    throw new IllegalStateException("openbanking_intent_id claim is missing from the access token");
+}
+
 // Get the intent Id from the request body
 def requestIntentId = request.entity.getJson().Data.ConsentId
 
 logger.debug(SCRIPT_NAME + 'Comparing token intent id {} with request intent id {}', accessTokenIntentId, requestIntentId)
 
 // Compare the id's and only allow the filter chain to proceed if they exists and they match
-if (accessTokenIntentId && requestIntentId && accessTokenIntentId == requestIntentId) {
+if (requestIntentId && accessTokenIntentId == requestIntentId) {
     // Request is valid, allow it to pass
     return next.handle(context, request)
 } else {
     Response response = new Response(Status.UNAUTHORIZED)
-    String message = 'invalid_request'
-    logger.error(SCRIPT_NAME + message + ' consentIds are not matching')
+    String message = 'consentId from the request does not match the openbanking_intent_id claim from the access token'
+    logger.error(SCRIPT_NAME + message)
     response.headers['Content-Type'] = 'application/json'
     response.entity = "{ \"error\":\"" + message + "\"}"
     return response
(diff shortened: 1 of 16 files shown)


mariantiris:
> 

Sure, we can do that, will have a look at everywhere we use attributes from the token and see if it's feasible to do this improvement.

mariantiris:
Found this already existing component: [SaveIntentIdOnAttributesContext](https://github.com/SecureApiGateway/secure-api-gateway-ob-uk/blob/master/config/7.1.0/securebanking/ig/scripts/groovy/SaveIntentIdOnAttributesContext.groovy) which parsed the openbanking_intent_id claim from the token and used it on all the routes where we needed the claim more than once to avoid parsing the JSON claims multiple times.

config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
```diff
@@ -0,0 +1,34 @@
...
+ * By doing this comparison, we prevent resources being submitted with a token obtained for another consent.
+ */
+
+String fapiInteractionId = request.getHeaders().getFirst(HttpHeaderNames.X_FAPI_INTERACTION_ID);
+if (fapiInteractionId == null) { fapiInteractionId = 'No ' + HttpHeaderNames.X_FAPI_INTERACTION_ID + ' header'}
+SCRIPT_NAME = '[ConsentIdValidator] (' + fapiInteractionId + ') - '
+logger.debug(SCRIPT_NAME + 'Running...')
+
+// Get the intent id from the access token
+def accessTokenIntentId = attributes.openbanking_intent_id
```
dbadham-fr, line 16: Can you add a check for this attribute being missing please? So that we can fail fast here.

If the attribute is missing, then we can assume that `SaveIntentIdOnAttributesContext` is missing or miss configured in the filter chain. This is therefore an application configuration issue.

Log a suitable error and throw an IllegalStateException (IG will then return a HTTP 500). This is how we handle such issues in the java filters.

config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
```diff
@@ -0,0 +1,34 @@
...
+if (accessTokenIntentId && requestIntentId && accessTokenIntentId == requestIntentId) {
+    // Request is valid, allow it to pass
+    return next.handle(context, request)
+} else {
+    Response response = new Response(Status.UNAUTHORIZED)
+    String message = 'invalid_request'
+    logger.error(SCRIPT_NAME + message + ' consentIds are not matching')
+    response.headers['Content-Type'] = 'application/json'
+    response.entity = "{ \"error\":\"" + message + "\"}"
+    return response
```
dbadham-fr, line 33: Can you trigger this in your environment please and add the error response json for review?

I think we can improve the message to state that the consentId in the request does not match the openbanking_intent_id in the access_token. 

It is good to be careful about leaking info in security related error cases, but if we do not mention the actual values then I think this is fine.

config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
mariantiris: Added the validation for the attribute. The behaviour is as follows:

An IllegalStateException error is thrown which result in this error exception in the logs:
`java.lang.IllegalStateException: openbanking_intent_id claim is missing from the access token`
and the caller will get this error response:

```json
{
    "Code": "500",
    "Id": "3b22668cdb2ab48ce91a1090079af85b",
    "Message": "[Status: 500 Internal Server Error]",
    "Errors": [
        {
            "ErrorCode": "UK.OBIE.UnexpectedError",
            "Message": "Internal error"
        }
    ]
}
```

config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
mariantiris: The error response looks as follows (updated with more details in the mesage as per the feedback):

```json
{
    "Code": "401",
    "Id": "fd9bd551f1d88428d90f82c2f620d58c",
    "Message": "[Status: 401 Unauthorized]",
    "Errors": [
        {
            "ErrorCode": "UK.OBIE.Unauthorized",
            "Message": "Unauthorized [consentId from the request does not match the openbanking_intent_id claim from the access token]"
        }
    ]
}
```

dbadham-fr: Looks good

config/7.1.0/securebanking/ig/scripts/groovy/ConsentIdValidator.groovy
mariantiris: Good point, updated the error message. Also added an error in the SaveIntentIdOnAttributesContext script.

config/7.1.0/securebanking/ig/scripts/groovy/SaveIntentIdOnAttributesContext.groovy
```diff
@@ -12,6 +12,11 @@ logger.debug(SCRIPT_NAME + "Running...")
 
 def intentId = JsonValue.json(Json.readJson(contexts.oauth2.accessToken.info.claims)).get("id_token").get("openbanking_intent_id").get("value").asString()
 logger.info(SCRIPT_NAME + "Intent Id value: " + intentId)
+
+if (!intentId) {
+    throw new IllegalStateException("openbanking_intent_id claim could not be extracted from the access token");
+}
```
dbadham-fr, line 18: This one needs a proper error message.

IllegalStateException (which in turn results in a 500) is reserved for situations where we have detected that the application has been configured incorrectly (or some programming error).

In this case, we have bad user input. We are expecting a token with some data in it but it's not there, so this should response with a 400 and a proper error msg.

mariantiris: Updated the error

dbadham-fr: Thanks

dbadham-fr (approved):

PR #386, opened 2023-05-17
578: Retain Path field in OB Errors object

dbadham-fr:
For validation errors, the RS will return Error objects which include a path field. The path indicates which field within the request is invalid.

Adding logic to retain this field if it is present in the RS response.

There's a separate issue to review and rewrite the ObResponseCheck script, see: https://github.com/SecureApiGateway/SecureApiGateway/issues/872

https://github.com/SecureApiGateway/SecureApiGateway/issues/578


dbadham-fr:
Example error response when required field is missing from a request:
```
{
    "Code": "400",
    "Id": "6c468b4c50c65ec0d9107cfb634fb1e4",
    "Message": "[Status: 400 Bad Request]",
    "Errors": [
        {
            "ErrorCode": "UK.OBIE.Field.Invalid",
            "Message": "The field received is invalid. Reason 'must not be null'",
            "Path": "data.psUAuthenticationMethod"
        }
    ]
}
```

jorgesanchezperez (approved):
Looks good to me

PR #390, opened 2023-06-08
1012: Domestic Payment Consent route to call RS

dbadham-fr:
Updating Domestic Payment Consent route to reverse proxy RS.

Removing filters which impl IDM Consent logic in IG from Domestic Payment routes.

Removing AM policy script call from Domestic Payment routes, this is now implemented in the RS/RCS.

Adding `x-api-client-id` header to all Domestic Payment API calls.

https://github.com/SecureApiGateway/SecureApiGateway/issues/1012


jorgesanchezperez (approved):
It's great to see a simplify the routes! nice job!

PR #400, opened 2023-07-07
1066: Consolidating ReverseProxyHandlers which call FR products

dbadham-fr:
We no longer need specific ReverseProxyHandlers for the RS, as the `x-ob-url` header is no longer required by any APIs (previously required by AISP). Once this is removed the handlers for FR products look identical.

Consolidating the handlers to: FRReverseProxyHandler and FRReverseProxyHandlerNoCapture

The OBReverseProxyHandler remains unchanged, and should be used when calling out to the OB Directory.

https://github.com/SecureApiGateway/SecureApiGateway/issues/1066

### DIFF
--- a/config/7.1.0/securebanking/ig/config/dev/config/config.json
+++ b/config/7.1.0/securebanking/ig/config/dev/config/config.json
@@ -172,19 +172,8 @@
       }
     },
     {
-      "name": "SBATReverseProxyHandlerIdentityPlatform",
-      "comment": "ReverseProxyHandler for calls to Identity Platform services (AM or IDM)",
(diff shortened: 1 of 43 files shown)


dbadham-fr:
On hold at the moment, the RS backoffice API for funds-confirmation requires the header. Once all of the payments have been moved to the new Consent Store, then this backoffice API will no longer be called.

All other func tests are passing.

dbadham-fr:
All payments have now been migrated, func tests are now passing.

PR ready for review.

jorgesanchezperez (approved):

PR #418, opened 2023-08-04
606: Funds confirmation availability route

jorgesanchezperez:
- Added route 67 for Funds confirmations
- Fix description of api client role check script filter

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/606


dbadham-fr (approved):

jorgesanchezperez:
> Route needs to add a filter to validate that the consentId in the payload matches the intent_id in the access token.
> 
> Take a look at the VRP funds confirmation route, it is using a filter to do this which can hopefully be reused.

Done

dbadham-fr (approved):

PR #421, opened 2023-08-11
968: Creating Filter to re-sign id_token JWS objects with the correct keyId

dbadham-fr:
This allows external clients to verify the signature of the id_token using an external trust directory, such as OpenBanking UK.

ReSignIdTokenFilter locates an id_token in an AM response and creates a new JWS using the configured signingKey and kid (which must match the trusted directory values)

Without this filter, AM will determine the kid using a proprietary algorithm which is unlikely to match the kid in the external jwks_uri, which therefore means that clients cannot verify the signature of the id_token (which is required for FAPI).

Relates to AM issue: https://bugster.forgerock.org/jira/browse/OPENAM-15617

https://github.com/SecureApiGateway/SecureApiGateway/issues/968


config/7.2.0/securebanking/ig/routes/routes-service/04-ob-as-authorize.json
```diff
@@ -36,7 +36,20 @@
...
           }
-        }        
+        },
+        {
+          "name": "ReSignIdTokenFilter",
+          "type": "ReSignIdTokenFilter",
+          "comment": "Re-sign the id_token returned by AM to fix OB keyId issue",
+          "config": {
+            "verificationSecretsProvider": "SecretsProvider-AmJWK",
+            "verificationSecretId": "any.value.in.regex.format",
```
dbadham-fr, line 46: same config used when we verify that `access_token`s have been signed by AM as part of route security.

BohoCode (approved):
Nice!

PR #436, opened 2023-10-06
898: Removing ParseCertificate groovy script

dbadham-fr:
ParseCertificate groovy script has become redundant now that we have `CertificateRetriever` java classes. 

New filter: `AddCertificateToAttributesContextFilter` has been created which can replace the ParseCertificate script in routes which need to access the cert multiple times. This new filter is responsible for retrieving the certificate from the header, parsing it and adding the X509Certificate object to the attributes context for other filters/scripts to retrieve. This means the certificate header is only parsed once, previously the DCR route was parsing the header twice.

A new CertificateRetriever impl: `ContextCertificateRetriever` has been created which fetches the certificate from the attributes context. This can be used to retrieve the certificate added to the context by the filter: `AddCertificateToAttributesContextFilter`

Adding Heaplets for both `CertificateRetriever` impls (`HeaderCertificateRetriever` and `ContextCertificateRetriever`)  so that they can be configured via the IG heap and plugged into filters as required.

Updating `FAPIAdvancedDCRValidationFilter`/`TransportCertValidationFilter`/`TokenEndpointTransportCertValidationFilter` configuration to allow the any CertificateRetriever to be used:
- Deprecating the "clientTlsCertHeader" option which is used to configure the HeaderCertificateRetriever
- New "certificateRetriever" option should be used in preference
  - This option allows greater flexibility, any retriever impl can be used by passing a heap reference
- This is a non-breaking change, the default config will continue to work

https://github.com/SecureApiGateway/SecureApiGateway/issues/898


dbadham-fr:
Converting this to a draft PR so that it is not accidentally merged until after the next release

jorgesanchezperez (commented):
Looks good to me!

dbadham-fr:
Rebased this PR after the 2.1.0 release, it is now ready for review

PR #437, opened 2023-10-09
Extendable signer

jorgesanchezperez:
- Draft of extendable signer solution
The platform needs to implement a feature to sign messages, for example to sign events.
This approach address the case where a customer could need implement a custom signer to sign messages, making the signer solution more easy to extend the solution.


secure-api-gateway-ig-extensions/src/main/java/com/forgerock/sapi/gateway/jws/signers/JwsSigner.java
```diff
@@ -0,0 +1,154 @@
...
+ *                 }, sapiJwsSignerException -> { // SapiJwsSignerException handler
+ *                     logger.error("Signature fails: {}", sapiJwsSignerException.getMessage())
+ *                     response.status = Status.INTERNAL_SERVER_ERROR
+ *                     response.entity = "{ \"error\":\"" + sapiJwsSignerException.getMessage() + "\"}"
+ *                     return newResultPromise(response)
+ *                 })
+ * }</pre>
+ * </p>
+ */
+public interface JwsSigner<T> {
```
jorgesanchezperez, line 125: Ignore the java doc at the moment, needs to be updated

PR #449, opened 2023-10-16
1151: Capturing Route Metrics

dbadham-fr:
Spec:
https://wikis.forgerock.org/confluence/pages/viewpage.action?spaceKey=SBAT&title=SAPI-G+Endpoint+Metrics

Adding filter: RouteMetricsFilter which is responsible for gathering and reporting metrics each time a route processes a request.

Installing the filter for the following routes:
- All RS OB routes
- The following AS routes
  - /register
  - /authorize
  - /access_token

Metrics are reported by publishing them to the log file, see LoggerRouteMetricsEventPublisher.
This has required changes to the logback config to create a custom appender for use by the publisher, this appender publishes a raw message with no additional formatting.

Example log:
```
{
    "timestamp": 1698157533193,
    "eventType": "route-metrics",
    "routeId": "04-ob-as-authorize",
    "requestPath": "/am/oauth2/realms/root/realms/alpha/authorize",
    "httpMethod": "POST",
    "responseTimeMillis": 258,
    "httpStatusCode": 302,
    "successResponse": true,
    "apiClientId": "910a246a-3778-442c-a525-2c21d8f5fd3c",
    "apiClientOrgId": "0015800001041REAAY",
    "trustedDirectory": "OpenBanking Ltd",
    "context": {}
}
```

Logging is done on a best effort basis. If a data point is missing then it will be logged as a null value. All data points should be available on the happy path. For error paths, the apiClient data may be missing if the client failed to authenticate.

To get the ApiClient data for all of the AS routes has required changes to filters in order to set the ApiClient into the attributes context, the FetchApiClientFilter is used for the RS routes but this does not work with the AS routes as it requires an access_token in the request entity which is not the case. Filters changed:
- CreateApiClient.groovy for /register
- TokenEndpointTransportCertValidationFilter for /access_token
- New filter: AuthorizeResponseFetchApiClientFilter for /authorize

The context field is optional in the metrics event, custom data is currently supplied only for the /access_token route which is used to identify the grant_type of the access_token requested. 

https://github.com/SecureApiGateway/SecureApiGateway/issues/1151

### DIFF
--- a/config/7.3.0/securebanking/ig/config/dev/config/logback.xml
+++ b/config/7.3.0/securebanking/ig/config/dev/config/logback.xml
@@ -25,34 +25,15 @@
             </layout>
         </encoder>
     </appender>
-     
-    <appender name="SIFT" class="ch.qos.logback.classic.sift.SiftingAppender">
-        <discriminator>
-            <key>routeId</key>
-            <defaultValue>system</defaultValue>
-        </discriminator>
-        <sift>
-            <!-- Create a separate log file for each <key> -->
-            <appender name="FILE-${routeId}" class="ch.qos.logback.core.rolling.RollingFileAppender">
-                <file>${openig.base}/logs/route-${routeId}.log</file>
-                 
-                <rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
-                    <!-- Rotate files daily -->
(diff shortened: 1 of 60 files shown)


jorgesanchezperez (approved):
Looks good to me, nice job!

BohoCode (approved):

PR #451, opened 2023-10-18
1162: store IDM orgName instead of orgId in api client org name

jorgesanchezperez:
Until now when a CDR is request, It was stored the `orgId` claim value from the Software Statement in the api client organisation name.
This is a fix to store the `org_name` claim value, retrieved from the Software Statement, in the api client organisation name instead of the `org_id` claim value.
- Added support to get the value from Software Statement `org_name` claim to be stored in IDM instead of orgId.
- Updated the trusted directories with a new property to retrieve and store the `org_name` claim value.
- Updated the current unit tests to cover the new `org_name` claim value retrieved from the Software Statement.

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/1162


jorgesanchezperez:
> Have you tried performing DCR and a flow using both Trusted Directories?

Yes, the image post in the above comment display two orgs, `Acme Fintech` org has been register using the SSA issued by Secure Api Gateway and `Forgerock Limited` has been register using the SSA issued by OBIE directory.

PR #453, opened 2023-10-30
1151: Cleanup IG rotue names to make metrics reports clearer

dbadham-fr:
The name of the route is used for the routeId in metrics events, renaming the following routes to make the metrics reports easier to understand:
- 54-ob-token-endpoint.json => 05-ob-as-token-endpoint.json , renaming to follow the same pattern as the authorize endpoint
- 60-internal-repo.json => 100-internal-repo.json , renaming due to clash with ob route 60-ob-file-payment-submission
- 99-ob-as.json => 99-ob-as-passthrough.json to make it clearer that requests are simply passed through to the AS.

https://github.com/SecureApiGateway/SecureApiGateway/issues/1151


jorgesanchezperez (approved):

BohoCode:
Is this a breaking change? 🤔

PR #467, opened 2023-12-06
961: Support OAuth 2.0 Pushed Authorization Requests (PAR)

dbadham-fr:
We want to be able to support Pushed Authorisation Requests as specified by https://datatracker.ietf.org/doc/html/rfc9126 . These requests should also be FAPI compliant, existing FAPI validation applied to `/authorize` requests also apply `/par`

Adding new IG route to reverse proxy the AM /par endpoint. This route is very similar to the existing /authorize route, it uses new PAR variants of existing Authorize filters.

`FapiAuthorizeRequestValidationFilter` has been refactored to enable it to be reused for /par request validation, base class:`BaseFapiAuthorizeRequestValidationFilter` has been extracted to handle the core logic of validating an authorize requests.

`FapiAuthorizeRequestValidationFilter` handles /authorize specific logic:
- handling Query param based parameters
- skipping validation if this is an /authorize request for a previously submitted /par request i.e. if the request_uri param is present

New `FapiParRequestValidationFilter` handles /par specific logic, which is to work with Form based parameters.

Refactored `AuthorizeResponseFetchApiClientFilter` to work with /par requests by allowing a Function to be plugged in which retrieves the client_id. This is used to enable the /authorize and /par routes to retrieve the ApiClient data to use in the metrics filter.

Created `ParResponseFetchApiClientFilterHeaplet` which an be used to create `ParResponseFetchApiClientFilter`s. This filter is an alias for AuthorizeResponseFetchApiClientFilter that has been configured to retrieve the client_id from the Request's Form.

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/964


BohoCode (commented):
https://github.com/SecureAPIGateway/SecureApiGateway/wiki/IG-Extensions-Java-Module needs updating to describe this filter?

BohoCode (approved):

dbadham-fr:
> https://github.com/SecureAPIGateway/SecureApiGateway/wiki/IG-Extensions-Java-Module needs updating to describe this filter?

Adding documentation now

PR #506, opened 2024-02-12
1279: Migrating FapiTransactionIdFilter groovy script to FapiInteractionIdTracingFilter java class

dbadham-fr:
FapiTransactionIdFilter.groovy has been ported to Java class: com.forgerock.sapi.gateway.fapi.FapiInteractionIdTracingFilter , this is a drop in replacement for the previous filter.

Minor impl change, the x-fapi-interaction-id is no longer written to the attributes context. Instead, if other filters need to get the value then they can retrieve it from the Request Headers. com.forgerock.sapi.gateway.fapi.FAPIUtils.getFapiInteractionId function has been updated to use the Request.

https://github.com/SecureApiGateway/SecureApiGateway/issues/1279


secure-api-gateway-ig-extensions/pom.xml
```diff
@@ -75,8 +75,8 @@
         </dependency>
 
         <dependency>
-            <groupId>org.slf4j</groupId>
-            <artifactId>slf4j-simple</artifactId>
+            <groupId>ch.qos.logback</groupId>
+            <artifactId>logback-classic</artifactId>
```
dbadham-fr, line 79: Required to test MDC contains the x-fapi-interaction-id. The slf4j-simple logger uses NOPMDCAdapter which does nothing when we attempt to add items to the MDC.

BohoCode (approved):

PR #519, opened 2024-03-14
Testing to see if we can set jwks_uri in oidc well known via AM config

BohoCode:


BohoCode:
This test proved the AM config can be used instead, so formalising this work in an issue 
https://github.com/SecureApiGateway/SecureApiGateway/issues/1318

PR #563, opened 2024-10-31
Conversion of error codes and messages.

mariantiris:
Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/1569

Added error code mappings in ObResponseCheck scripts - this is responsible of creating the OB compliant error responses that are thrown directly from IG.

### DIFF
--- a/config/7.3.0/securebanking/ig/scripts/groovy/ObResponseCheck.groovy
+++ b/config/7.3.0/securebanking/ig/scripts/groovy/ObResponseCheck.groovy
@@ -22,7 +22,7 @@ SCRIPT_NAME = "[ObResponseCheck] (" + fapiInteractionId + ") - ";
 logger.debug(SCRIPT_NAME + "Running...")
 
 String HEADER_INTERACTION_ID = "x-fapi-interaction-id"
-Map<String, String> getGenericError(Status status, String responseBody) {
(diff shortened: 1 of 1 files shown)


config/7.3.0/securebanking/ig/scripts/groovy/ObResponseCheck.groovy
```diff
@@ -99,6 +99,21 @@ static isObCompliantError(responseBody) {
...
+  String apiVersionRegex = "(v\\d+)"
+  def match = (request.uri.path =~ apiVersionRegex)
+  def isV3Request = true
+  if (match.find()) {
+    def apiVersion = match.group(1)  // Capture only the major version, e.g., "v4"
+    if (apiVersion == "v4") {
+      return false
+    }
+  }
+  return true;
```
dbadham-fr, line 112: Can you rewrite this not to use a regex please?

The groovy `=~` operator is basically syntactic sugar for `java.util.regex.Pattern.compile(string)` which has a relatively high performance cost, and one compile call is being made each time is this method is called.

As the path element containing the version is constant, then its more efficient to manually parse the String. The `org.forgerock.http.MutableUri#getPathElements` method of request.uri field is handy to get the path elements as an array.

dbadham-fr (approved):

PR #570, opened 2024-12-12
1420: Fix config for IG 2024.11 with openig-fapi

BohoCode:
Using the config in core as a basis. Also changed AmService-OBIE to just AmService

Issue: https://github.com/SecureApiGateway/SecureApiGateway/issues/1620


shaunharrisonFR:
Actions by Shaun and Wayne since the PR originally got created

- Merge Master to bump Parent & Core versions
- Separate to this PR, we've deployed the TTD as standalone
- Removed 7# routes as now handled in TTD
- Removed what we've decided is an unneeded filter from route 03

wmorrison365fr:
Superseded (and tidied up) by PR #574

PR #617, opened 2025-09-15
OPENIG-9515 Introduce RCS API client error tranform filter

wmorrison365fr:
OPENIG-9515 Introduce error transform filter in route 83-rcs-api to ensure client errors are correctly reported according to OAuth2 standards.


wmorrison365fr:
Change not required. Underlying issue found in RCS:
https://github.com/SecureApiGateway/secure-api-gateway-ob-uk-rcs/pull/305/files

PR #680, opened 2026-06-11
OPENIG-10460 Fix logback by unpinning and using IG BOM config

wmorrison365fr:


guillaumelamirand (approved):

wmorrison365fr:
Local build env is broken (as component). 
Tested using env dev-cdk-ob instead.

